PR #25, opened 2024-06-20
feat: server settings

danieldietzler:


marcelfloe (changes requested):
JavaDoc missing

server/src/main/java/de/uftos/controller/ServerController.java
```diff
@@ -2,26 +2,59 @@
...
 @RestController
 @RequestMapping("/server")
 public class ServerController {
-  private ServerService serverService;
+  private final ServerService serverService;
 
+  /**
+   * Creates the server controller.
+   *
+   * @param serverService the service for the student entity.
```
marcelfloe, line 23: student entity -> server entity

server/src/main/java/de/uftos/controller/ServerController.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+@RestController
+@RequestMapping("/server")
+public class ServerController {
+  private final ServerService serverService;
+
+  /**
+   * Creates the server controller.
+   *
+   * @param serverService the service for the student entity.
```
danieldietzler, line 23:
```suggestion
   * @param serverService the service for the server entity.
```

marcelfloe (approved):

PR #97, opened 2024-06-30
Feat/teacher service todos

Lama64:


server/src/main/java/de/uftos/services/TeacherService.java
```diff
@@ -44,7 +44,30 @@ public TeacherService(TeacherRepository repository) {
...
+    lastName.ifPresent((param) -> spec.or(createFilter(param, "lastName")));
+    acronym.ifPresent((param) -> spec.or(createFilter(param, "acronym")));
+    subjects.ifPresent((param) -> {
+      for (String subjectId : param) {
+        spec.and(dropDownFilter(subjectId, "subjects"));
+      }
+    });
+    tags.ifPresent((param) -> {
+      for (String tagId : param) {
+        spec.and(dropDownFilter(tagId, "tags"));
```
SuperMarcomen, line 58: You should write spec = spec.and... because spec.and returns a new Specification instance. Specification is just like String, immutable. This should be changed everywhere where you have spec.or or spec.and (And therefore you can not use a function needing a lamba like ifPresent

danieldietzler (approved):

PR #101, opened 2024-07-01
feat: constraint instances

### LINKED ISSUE #46: server: constraints

danieldietzler:
Closes #46


server/src/main/java/de/uftos/builders/SpecificationBuilder.java
```diff
@@ -64,8 +67,41 @@ public SpecificationBuilder<T> optionalAndJoinIn(Optional<String[]> attributeVal
     }
     specification = specification.and(
         (root, query, cb) -> root.join(relationName).get(attributeName)
-            .in(List.of(attributeValue.get()))
+            .in((Object[]) attributeValue.get())
```
danieldietzler, line 70: Why does this need to be casted to an `Object[]`?

server/src/main/java/de/uftos/builders/SpecificationBuilder.java
```diff
@@ -64,8 +67,41 @@ public SpecificationBuilder<T> optionalAndJoinIn(Optional<String[]> attributeVal
...
+    if (attributeValue.isEmpty()) {
+      return this;
+    }
+    specification = specification.and(
+        (root, query, cb) -> cb.equal(root.join(relationName).get(attributeName), attributeValue)
+    );
+    return this;
+  }
+
+  private Predicate equalsIgnoreCase(CriteriaBuilder cb, Expression<String> value, String pattern) {
```
danieldietzler, line 104: This is `likeIgnoreCase`

danieldietzler: Might as well call it `iLike`

server/src/main/java/de/uftos/services/ConstraintInstanceService.java
```diff
@@ -36,7 +86,36 @@ public ConstraintInstanceService(ConstraintInstanceRepository repository) {
...
+
+    if (signature.getParameters().size() != request.arguments().size()) {
+      throw new ResponseStatusException(HttpStatus.BAD_REQUEST);
+    }
+
+    for (ConstraintParameter parameter : signature.getParameters()) {
+      String value = request.arguments().get(parameter.getParameterName());
+      if (value == null || value.isEmpty()) {
+        throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
+            "Parameter %s with could not be found".formatted(parameter.getParameterName()));
```
danieldietzler, line 100:
```suggestion
            "Parameter %s could not be found".formatted(parameter.getParameterName()));
```

server/src/main/java/de/uftos/builders/SpecificationBuilder.java
SuperMarcomen: Before I was passing a List of `String[]` which is wrong. Now I'm passing an array of `Object`. The cast was needed because the call to the method `in` was ambigous and it was not clear whether the varargs a non-varargs call was desired. I have since replaced it with something better.

server/src/main/java/de/uftos/builders/SpecificationBuilder.java
SuperMarcomen: Thanks, done.

server/src/main/java/de/uftos/utils/SpecificationBuilder.java
```diff
@@ -64,8 +68,41 @@ public SpecificationBuilder<T> optionalAndJoinIn(Optional<String[]> attributeVal
...
     return this;
   }
+
+  public SpecificationBuilder<T> andEqual(String attributeValue, String attribute) {
+    specification =
+        specification.and((root, query, cb) -> cb.equal(root.get(attribute), attributeValue));
+    return this;
+  }
+
+  public SpecificationBuilder<T> optionalAndJoinLike(Optional<String> attributeValue,
```
danieldietzler, line 82: This should also be called `optionalAndJoinLikeIgnoreCase` or `optionalAndJoinILike`

danieldietzler: @SuperMarcomen if this has been resolved I'm happy to merge it

SuperMarcomen: done

SuperMarcomen (approved):

PR #105, opened 2024-07-02
fix: database relations

danieldietzler:
To remove orphaned tables the DB volume should be deleted (`docker compose down -v db`)
Screenshot 2024-07-02 at 16 56 27


SuperMarcomen (approved):
LGTM!

PR #106, opened 2024-07-02
feat:students service

SuperMarcomen:


server/src/main/java/de/uftos/services/StudentService.java
```diff
@@ -39,7 +40,28 @@ public StudentService(StudentRepository repository) {
...
+  private Specification<Student> createSpecification(Optional<String> firstName,
+                                                         Optional<String> lastName,
+                                                         Optional<String[]> tags) {
+    Specification<Student> spec = Specification.where(null);
+
+    firstName.ifPresent(param -> spec.or(createFilter(param, "firstName")));
+    lastName.ifPresent(param -> spec.or(createFilter(param, "lastName")));
+    tags.ifPresent(param -> {
+      for (String tag : param) {
+        spec.or(createFilter(tag, "tags"));
```
Lama64, line 56: tags filter should be connected with and, because tags are filtered with a dropdown menu and only students should be shown which have all selected tags.
CreateFilter method not usable for tags, as ids of tags are given to filter.
A hopefully working version of a filter for tags should soon be available in PR #97.

Lama64 (changes requested):

server/src/main/java/de/uftos/services/StudentService.java
```diff
@@ -39,7 +40,28 @@ public StudentService(StudentRepository repository) {
...
                            Optional<String> lastName, Optional<String[]> tags) {
-    return this.repository.findAll(pageable);
+    Specification<Student> spec = createSpecification(firstName, lastName, tags);
+    return this.repository.findAll(spec, pageable);
+  }
+
+  private Specification<Student> createSpecification(Optional<String> firstName,
+                                                         Optional<String> lastName,
+                                                         Optional<String[]> tags) {
+    Specification<Student> spec = Specification.where(null);
```
Lama64, line 50: null as a parameter, might be dangerous, was this tested?
Using cb.disjunction() as an always false might be a better option.

SuperMarcomen: Yes, it was tested. where(null) is also always false.

server/src/main/java/de/uftos/builders/SpecificationBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+
+  public SpecificationBuilder() {
+    specification = Specification.where(null);
+  }
+
+  public Specification<T> build() {
+    return specification;
+  }
+
+  public SpecificationBuilder<T> addOptionalOr(Optional<String> param, String paramName) {
```
danieldietzler, line 19: IMO the `add` can be omitted here, so just `optionalOr` (same for the other methods)

server/src/main/java/de/uftos/builders/SpecificationBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+  private Specification<T> createOrFilter(String param, String paramName) {
+    return ((root, query, cb) ->  cb.like(root.get(paramName), "%" + param + "%"));
+  }
+
+  private Specification<T> createJoinFilter(String relationName, String attributeName, List<String> attributeValues) {
+    return (root, query, cb) -> {
+      // Create a predicate to check if the tag ID matches
+      return root.join(relationName).get(attributeName).in(attributeValues);
+    };
+  }
```
danieldietzler, line 40: Those methods don't really bring any benefit imo

server/src/main/java/de/uftos/builders/SpecificationBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+    return specification;
+  }
+
+  public SpecificationBuilder<T> addOptionalOr(Optional<String> param, String paramName) {
+    if (param.isEmpty()) return this;
+    specification = specification.or(createOrFilter(param.get(), paramName));
+    return this;
+  }
+
+  public SpecificationBuilder<T> addOptionalJoinFilter(Optional<String[]> attributeValue, String relationName, String attributeName) {
```
danieldietzler, line 25: This one and the method above imply (by their name) that they are very generic, while in reality they specifically do LIKE and IN. Certainly there will be more methods necessary in the future, so it may be worth considering how to make those more generic

server/src/main/resources/application.properties
```diff
@@ -1,9 +1,9 @@
 spring.application.name=uftos
 spring.devtools.livereload.enabled=true
 spring.devtools.restart.enabled=true
-spring.datasource.url=jdbc:postgresql://${DB_HOST}:${DB_PORT}/${DB_NAME}
-spring.datasource.username=${DB_USER}
-spring.datasource.password=${DB_PASSWORD}
+spring.datasource.url=jdbc:postgresql://${DB_HOST:localhost}:${DB_PORT:5432}/${DB_NAME:uftos}
+spring.datasource.username=${DB_USER:uftos}
+spring.datasource.password=${DB_PASSWORD:superSecurePassword}
```
danieldietzler, line 6: Those default values should be undone

server/src/main/java/de/uftos/builders/SpecificationBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+    return this;
+  }
+
+  private Specification<T> createOrFilter(String param, String paramName) {
+    return ((root, query, cb) ->  cb.like(root.get(paramName), "%" + param + "%"));
+  }
+
+  private Specification<T> createJoinFilter(String relationName, String attributeName, List<String> attributeValues) {
+    return (root, query, cb) -> {
+      // Create a predicate to check if the tag ID matches
```
danieldietzler, line 37: This comment doesn't add any value
```suggestion
```

server/src/test/java/de/uftos/controller/StudentControllerTest.java
```diff
@@ -0,0 +1,154 @@
...
+  static String tagId;
+
+  @BeforeAll
+  static void createTestStudents() {
+    tagId = given().contentType(ContentType.JSON)
+        .body("""
+                    {
+                      "tagName": "Sehebehinderung"
+                    }""")
+        .when()
```
danieldietzler, line 26: Instead of having incredibly ugly multiline strings, can we use some JSON builder?

server/src/main/java/de/uftos/builders/SpecificationBuilder.java
SuperMarcomen: Not knowing what our feature needs will be, I think it would be quite hard to generalize this method right now.

server/src/main/java/de/uftos/builders/SpecificationBuilder.java
SuperMarcomen: Yes, you're right. This is the result of copy/pasting the code from a place where these methods where actually useful

server/src/main/resources/application.properties
SuperMarcomen: done

server/src/main/java/de/uftos/builders/SpecificationBuilder.java
SuperMarcomen: removed

server/src/main/java/de/uftos/builders/SpecificationBuilder.java
SuperMarcomen: Now that I think of it, I like your proprosition better

server/src/main/java/de/uftos/builders/SpecificationBuilder.java
danieldietzler: Then we should probably name it less generic.

SuperMarcomen: done

server/src/test/java/de/uftos/controller/StudentControllerTest.java
SuperMarcomen: done

server/src/test/java/de/uftos/controller/StudentControllerTest.java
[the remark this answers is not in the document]
SuperMarcomen: makes sense, done

server/src/main/java/de/uftos/services/StudentService.java
SuperMarcomen: I tested your proposition and it didn't work, but mine didn't either. Daniel and I found the correct way of doing it

Lama64 (approved):

server/src/test/java/de/uftos/e2e/StudentsTest.java
```diff
@@ -0,0 +1,157 @@
...
+        .get("/students")
+        .then()
+        .statusCode(200)
+        .body("totalElements", equalTo(1))
+        .body("content[0].id", equalTo(secondStudent))
+        .log().ifValidationFails();
+  }
+
+  @Test
+  void getStudentsWithNameTest() throws JSONException {
```
danieldietzler, line 144:
```suggestion
  void getAllStudents() throws JSONException {
    given().contentType(ContentType.JSON)
        .body(generatePageJson(0, 10, Collections.emptyList()))
        .when()
        .get("/students")
        .then()
        .statusCode(200)
        .body("totalElements", equalTo(2))
        .log().ifValidationFails();
  }

  @Test
  void getStudentsWithTag() throws JSONException {
    given().contentType(ContentType.JSON)
        .body(generatePageJson(0, 10, Collections.emptyList()))
        .param("tags", List.of(tagId))
        .when()
        .get("/students")
        .then()
        .statusCode(200)
        .body("totalElements", equalTo(1))
        .body("content[0].id", equalTo(secondStudent))
        .log().ifValidationFails();
  }

  @Test
  void getStudentsWithName() throws JSONException {
```

danieldietzler (approved):

PR #125, opened 2024-07-07
Data Table

Lama64:
admin/rooms for example of formatting with static data.
admin/students for database integration (reload of page needed after using add button).

### DIFF
--- a/web/src/lib/elements/ui/dataTable/data-table-actions.svelte
+++ b/web/src/lib/elements/ui/dataTable/data-table-actions.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
   import Ellipsis from 'lucide-svelte/icons/ellipsis';
   import * as DropdownMenu from '$lib/elements/ui/dropdown-menu';
-  import * as Button from '$lib/elements/ui/button';
+  import Button from '$lib/elements/ui/button/button.svelte';
 
   export let id: string;
 </script>
--- a/web/src/lib/elements/ui/dataTable/data-table.svelte
+++ b/web/src/lib/elements/ui/dataTable/data-table.svelte
@@ -11,11 +11,11 @@
     addSelectedRows,
   } from 'svelte-headless-table/plugins';
   import { Button } from '$lib/elements/ui/button';
-  import ArrowUpDown from 'lucide-svelte/icons/arrow-up-down';
   import ChevronDown from 'lucide-svelte/icons/chevron-down';
-  import * as Input from '$lib/elements/ui/input';
   import * as DropdownMenu from '$lib/elements/ui/dropdown-menu';
   import DataTableCheckbox from './data-table-checkbox.svelte';
+  import { ArrowDown, ArrowUp } from 'lucide-svelte';
+  import Input from '$lib/elements/ui/input/input.svelte';
 
   interface DataItem {
     id: string;
@@ -26,7 +26,9 @@
   export let data: DataItem[];
   export let columnNames;
 
-  const table = createTable(writable(data), {
+  let tableData = writable(data);
+
+  const table = createTable(tableData, {
     page: addPagination(),
     sort: addSortBy(),
     filter: addTableFilter({
@@ -37,12 +39,11 @@
   });
 
   let idKey = Object.keys(data[0])[0];
-  console.log(idKey);
   let columns = table.createColumns([
     table.column({
       //first column only contains the checkboxes.
       accessor: (item) => {
-        return item[idKey]; //'item' is of type 'unknown' error dont know how to fix
+        return item[idKey];
       },
       id: 'id',
       header: (_, { pluginStates }) => {
@@ -114,7 +115,7 @@
     .filter(([, hide]) => !hide)
     .map(([id]) => id);
 
-  const hidableCols = ['status', 'email', 'amount'];
+  const hidableCols = columnNames;
 </script>
 
 <div>
@@ -146,11 +147,14 @@
             <Table.Row>
               {#each headerRow.cells as cell (cell.id)}
                 <Subscribe attrs={cell.attrs()} let:attrs props={cell.props()} let:props>
-                  <Table.Head {...attrs} class="[&:has([role=checkbox])]:pl-3">
+                  <Table.Head {...attrs} class="[&:has([role=checkbox])]:pl-3 text-white bg-foreground">
                     {#if cell.id !== 'actions' && cell.id !== 'id'}
                       <Button variant="ghost" on:click={props.sort.toggle}>
                         <Render of={cell.render()} />
-                        <ArrowUpDown class={'ml-2 h-4 w-4'} />
+                        <div class="flex ml-2">
+                          <ArrowUp class="h-4 w-4 {props.sort.order === 'desc' ? 'text-accent' : ''}" />
+                          <ArrowDown class="ml-[-4px] h-4 w-4 {props.sort.order === 'asc' ? 'text-accent' : ''}" />
+                        </div>
                       </Button>
                     {:else}
                       <Render of={cell.render()} />
@@ -168,7 +172,7 @@
             <Table.Row {...rowAttrs} {...rowAttrs} data-state={$selectedDataIds[row.id] && 'selected'}>
               {#each row.cells as cell (cell.id)}
                 <Subscribe attrs={cell.attrs()} let:attrs>
-                  <Table.Cell {...attrs}>
+                  <Table.Cell {...attrs} class="bg-white">
                     <Render of={cell.render()} />
                   </Table.Cell>
                 </Subscribe>
--- a/web/src/lib/elements/ui/dataTable/data-table.svelte
+++ b/web/src/lib/elements/ui/dataTable/data-table.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
   import { createTable, Render, Subscribe, createRender } from 'svelte-headless-table';
-  import { writable } from 'svelte/store';
   import * as Table from '$lib/elements/ui/table';
   import DataTableActions from './data-table-actions.svelte';
   import {
@@ -14,6 +13,7 @@
   import ChevronDown from 'lucide-svelte/icons/chevron-down';
   import * as DropdownMenu from '$lib/elements/ui/dropdown-menu';
   import DataTableCheckbox from './data-table-checkbox.svelte';
+  import DataTableHeaderCheckbox from './data-table-header-checkbox.svelte';
   import { ArrowDown, ArrowUp } from 'lucide-svelte';
   import Input from '$lib/elements/ui/input/input.svelte';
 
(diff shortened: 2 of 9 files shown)


SuperMarcomen (approved):

web/src/lib/elements/ui/dataTable/data-table-actions.svelte
```diff
@@ -15,11 +19,17 @@
   </DropdownMenu.Trigger>
   <DropdownMenu.Content>
     <DropdownMenu.Group>
-      <DropdownMenu.Label>Actions</DropdownMenu.Label>
-      <DropdownMenu.Item on:click={() => navigator.clipboard.writeText(id)}>Copy payment ID</DropdownMenu.Item>
+      <DropdownMenu.Label>Aktionen</DropdownMenu.Label>
+      <DropdownMenu.Item on:click={() => navigator.clipboard.writeText(id)}>ID kopieren</DropdownMenu.Item>
     </DropdownMenu.Group>
+    <DropdownMenu.Item on:click={() => goto($page.url + '/' + id)}>Editieren</DropdownMenu.Item>
```
danieldietzler, line 25:
```suggestion
    <DropdownMenu.Item on:click={() => goto(`${$page.url}/${id}`)}>Editieren</DropdownMenu.Item>
```

web/src/lib/elements/ui/dataTable/data-table.svelte
```diff
@@ -100,12 +106,22 @@
...
   );
 
-  const { headerRows, pageRows, tableAttrs, tableBodyAttrs, pluginStates, flatColumns, rows } =
-    table.createViewModel(columns);
+  let tableOptions = {
+    rowDataId: (item: { [x: string]: any }, index: number) => {
+      if (item) {
+        return item['id'];
+      } else return index;
+    },
```
danieldietzler, line 114: This can be simplified to `=> item ? item['id']  : index

web/src/lib/elements/ui/dataTable/data-table.svelte
```diff
@@ -114,15 +130,47 @@
...
     .map(([id]) => id);
 
-  const hidableCols = ['status', 'email', 'amount'];
+  const hidableCols = keys.slice(1);
+
+  async function getData() {
+    let sortKey: SortKey = $sortKeys[0];
+    let sortString;
+    if (sortKey) {
+      sortString = sortKey.id + ',' + sortKey.order;
```
danieldietzler, line 139:
```suggestion
      sortString = `${sortKey.id},${sortKey.order}`;
```
Template string are awesome! You should use them :)

web/src/lib/elements/ui/dataTable/data-table.svelte
```diff
@@ -114,15 +130,47 @@
     .filter(([, hide]) => !hide)
     .map(([id]) => id);
 
-  const hidableCols = ['status', 'email', 'amount'];
+  const hidableCols = keys.slice(1);
+
+  async function getData() {
+    let sortKey: SortKey = $sortKeys[0];
+    let sortString;
+    if (sortKey) {
```
danieldietzler, line 138: This can again be simplified as a ternary.

web/src/lib/elements/ui/dataTable/data-table.svelte
```diff
@@ -114,15 +130,47 @@
...
+  async function onDeleteKey(e: KeyboardEvent) {
+    if (e.code !== 'Delete') {
+      return;
+    }
+    console.log($selectedDataIds);
+    let promises: Promise<void>[] = [];
+    for (let i = 0; i < Object.keys($selectedDataIds).length; i++) {
+      let rowToDelete = Object.keys($selectedDataIds)[i];
+      promises.push(deleteEntry(rowToDelete));
+    }
```
danieldietzler, line 153: This can be a for each

web/src/lib/elements/ui/dataTable/data-table.svelte
```diff
@@ -114,15 +130,47 @@
...
+    await Promise.all(promises);
```
danieldietzler, line 154: Combining them all is a good performance improvement!

web/src/lib/elements/ui/dataTable/data-table.svelte
```diff
@@ -114,15 +130,47 @@
...
+      sortString = sortKey.id + ',' + sortKey.order;
+    } else sortString = '';
+    await loadPage($pageIndex, sortString, $filterValue);
+  }
+
+  async function onDeleteKey(e: KeyboardEvent) {
+    if (e.code !== 'Delete') {
+      return;
+    }
+    console.log($selectedDataIds);
```
danieldietzler, line 148: This should be removed, right?

web/src/lib/elements/ui/dataTable/data-table.svelte
```diff
@@ -184,10 +242,24 @@
       {Object.keys($selectedDataIds).length} von{' '}
       {$rows.length} Zeile(n) ausgewählt.
     </div>
-    <Button disabled={!$hasPreviousPage} on:click={() => ($pageIndex = $pageIndex - 1)} size="sm" variant="outline"
+    <Button
+      disabled={!$hasPreviousPage}
+      on:click={async () => {
+        $pageIndex = $pageIndex - 1;
```
danieldietzler, line 248: Can this be simplified to `$pageIndex--` or is the reactivity broken then?

web/src/lib/elements/ui/dataTable/data-table.svelte
```diff
@@ -184,10 +242,24 @@
...
+        await getData();
```
danieldietzler, line 249: Since you'll need to get the data whenever the pageIndex changes you should consider using something like `$: $pageIndex, await getData()`

web/src/routes/admin/(resources)/rooms/+page.svelte
```diff
@@ -1,25 +1,49 @@
...
+  let tableData: Writable<DataItem[]> = writable([]);
+  let totalElements: Writable<number> = writable(0);
+
+  onMount(async () => await loadPage(0, '', ''));
+
+  async function loadPage(index: number, sortString: string, filter: string) {
+    let pageable: Pageable = { page: index, size: 10, sort: [sortString] };
+    let result: PageRoom;
+    try {
+      /^\d+$/.test(filter)
```
danieldietzler, line 19: WTH? You're checking if the filter only consists of digits? Why?

web/src/routes/admin/(resources)/rooms/+page.svelte
```diff
@@ -1,25 +1,49 @@
...
+        ? (result = await getRooms(pageable, {
+            name: filter,
+            buildingName: filter,
+            capacity: filter as unknown as number,
+          }))
+        : (result = await getRooms(pageable, {
+            name: filter,
+            buildingName: filter,
+          }));
+      totalElements.set(result.totalElements as number);
```
danieldietzler, line 29: Actually converting it (`Number(result.totalElements)`) is cleaner here

web/src/routes/admin/(resources)/rooms/+page.svelte
```diff
@@ -1,25 +1,49 @@
...
+      tableData.set(result.content as unknown as DataItem[]);
+    } catch {
+      error(404, { message: 'Could not fetch page' });
```
danieldietzler, line 32: Not sure if a 404 is fitting here. What were your thoughts?

web/src/routes/admin/(resources)/students/+page.svelte
```diff
@@ -0,0 +1,42 @@
...
+  import DataTable, { type DataItem } from '$lib/elements/ui/dataTable/data-table.svelte';
+  import { deleteStudent, getStudents, type Pageable, type PageStudent } from '$lib/sdk/fetch-client';
+  import { error } from '@sveltejs/kit';
+  import { onMount } from 'svelte';
+  import { writable, type Writable } from 'svelte/store';
+
+  let columnNames = ['Vorname', 'Nachname', 'Tags'];
+  let keys = ['id', 'firstName', 'lastName', 'tags'];
+  let tableData: Writable<DataItem[]> = writable();
+  let totalElements: Writable<number> = writable(0);
```
danieldietzler, line 11: This whole store handling is honestly super weird. `loadPage` should just return the page and then you can do whatever with it.

web/src/routes/admin/(resources)/students/+page.svelte
```diff
@@ -0,0 +1,42 @@
...
+    } catch {
+      error(404, { message: 'Could not fetch page' });
+    }
+  }
+
+  async function deleteEntry(id: string) {
+    try {
+      await deleteStudent(id);
+    } catch {
+      error(400, { message: `student with id ${id} could not be found` });
```
danieldietzler, line 33: In this case you could argue that it's a 404

web/src/routes/admin/(resources)/students/+page.svelte
```diff
@@ -0,0 +1,42 @@
...
+
+  async function loadPage(index: number, sortString: string, filter: string) {
+    let pageable: Pageable = { page: index, size: 10, sort: [sortString] };
+    try {
+      const result: PageStudent = await getStudents(pageable, {
+        firstName: filter,
+        lastName: filter,
+      });
+      totalElements.set(result.totalElements as number);
+      tableData.set(result.content as unknown as DataItem[]);
```
danieldietzler, line 23: Why do you have to cast this to unknown first? I'd expect `StudentResponseDto` and `DataItem` to have sufficient overlap. If not you should really just write a mapping function instead of doing this

web/src/routes/admin/(resources)/subjects/+page.ts
```diff
@@ -3,7 +3,7 @@ import type { PageLoad } from './$types';
 export const load = (() => {
   return {
     meta: {
-      title: 'Subjects',
+      title: 'Fachmanagment',
```
danieldietzler, line 6: Thoughts on just using "Fächer" instead of adding "management" to everything?

danieldietzler (commented):
Is it possible you're just rolling a dice to figure out HTTP codes? lmao

web/src/routes/admin/(resources)/rooms/+page.svelte
Lama64: Why wouldn't 404 be fitting here? What error code would you choose instead?

web/src/routes/admin/(resources)/students/+page.svelte
Lama64: Found a way to avoid these stores, but could you take a look at the comment at the bottom of the students file where the data table component is. Maybe you have a better idea.

web/src/routes/admin/(resources)/students/+page.svelte
Lama64: Changed the message to better fit the 400 error code, to resemble the other errors with code 400.

danieldietzler: That's generally an interesting debate. You *can* argue that — since `id` is a path variable — accessing `/students/:id` results in a 404 page not found, as that path doesn't exist if the id doesn't exist. However, since we're not talking about static paths here, some people also say that a 404 is inappropriate.

PR #128, opened 2024-07-09
feat: add color for subjects

danieldietzler:


server/src/main/java/de/uftos/entities/Subject.java
```diff
@@ -33,6 +33,8 @@ public class Subject {
   @NotEmpty
   private String name;
 
+  private String color;
```
SuperMarcomen, line 36: You should also update the the `SubjectRequestDto`

SuperMarcomen (changes requested):

server/src/main/java/de/uftos/entities/Subject.java
danieldietzler: Oh true, sorry

SuperMarcomen (approved):

PR #132, opened 2024-07-11
Parser

### LINKED ISSUE #44: server: parser

marcelfloe:
Closes #44


server/src/main/java/de/uftos/dto/ucdl/ast/ControlSequenceDto.java
danieldietzler, line 1: Could you elaborate what all these abstrac syntax trees containing x do and why we need some specific ones while others are covered by the generic AST?

server/src/main/java/de/uftos/repositories/ucdl/UcdlRepositoryImpl.java
```diff
@@ -11,24 +17,63 @@
...
+      int readInformation = reader.read();
+      StringBuilder sb = new StringBuilder();
+      while (readInformation >= 0) {
+        sb.append((char) readInformation);
+        readInformation = reader.read();
+      }
+      return sb.toString();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
```
danieldietzler, line 37: Could we use the more modern implementation from the `nio` package?

server/src/main/java/de/uftos/repositories/ucdl/UcdlRepositoryImpl.java
```diff
@@ -11,24 +17,63 @@
...
+      FileReader reader = new FileReader(ucdlFile);
+      int readInformation = reader.read();
+      StringBuilder sb = new StringBuilder();
+      while (readInformation >= 0) {
+        sb.append((char) readInformation);
+        readInformation = reader.read();
+      }
+      return sb.toString();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
danieldietzler, line 36: We shouldn't throw generic runtime exceptions when interacting with the file.

server/src/main/java/de/uftos/repositories/ucdl/UcdlRepositoryImpl.java
```diff
@@ -11,24 +17,63 @@
...
+      return new ParsingResponse(false, e.getMessage());
+    }
+    return new ParsingResponse(true, "Parsing was successful!");
   }
 
   @Override
-  public HashMap<String, ConstraintDefinitionDto> getConstraints() throws ParseException {
-    return null;
+  public HashMap<String, ConstraintDefinitionDto> getConstraints() {
+    if (currentDefinitions == null) {
```
danieldietzler, line 65:
```suggestion
    if (this.currentDefinitions == null) {
```

server/src/main/java/de/uftos/repositories/ucdl/UcdlRepositoryImpl.java
```diff
@@ -11,24 +17,63 @@
...
+    } catch (ParseException | IOException e) {
+      return new ParsingResponse(false, e.getMessage());
+    }
+    return new ParsingResponse(true, "Parsing was successful!");
   }
 
   @Override
-  public HashMap<String, ConstraintDefinitionDto> getConstraints() throws ParseException {
-    return null;
+  public HashMap<String, ConstraintDefinitionDto> getConstraints() {
```
danieldietzler, line 64: You could simplify this whole function to
```java
if (this.currentDefinitions == null && !this.parseFile().success()) {
	return null;
}
return this.currentDefinitions;
```
Then you can get rid of the `setCurrentDefinitions()` function as well

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,685 @@
...
+import de.uftos.dto.ucdl.ast.ValueDto;
+import de.uftos.repositories.ucdl.parser.javacc.Node;
+import de.uftos.repositories.ucdl.parser.javacc.ParseException;
+import de.uftos.repositories.ucdl.parser.javacc.SimpleNode;
+import de.uftos.repositories.ucdl.parser.javacc.SyntaxChecker;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+//TODO: !!!REFACTOR!!!
```
danieldietzler, line 20: lmao

server/src/main/java/de/uftos/repositories/ucdl/parser/UcdlParser.java
```diff
@@ -0,0 +1,119 @@
...
+              + " \"SOFT_PENALIZE\", or empty field expected!");
+    };
+
+    HashMap<String, ResourceType> parameters = new HashMap<>();
+    constraintDefinition.get("parameter").fields().forEachRemaining((entry) -> {
+      if (entry != null) {
+        try {
+          parameters.put(entry.getKey(), getResourceType(entry.getValue().textValue()));
+        } catch (ParseException e) {
+          throw new RuntimeException(e);
```
danieldietzler, line 86: A general runtime exception is bad here

server/src/main/java/de/uftos/repositories/ucdl/parser/UcdlParser.java
```diff
@@ -0,0 +1,119 @@
...
+        parseDefinition(constraintDefinition.get("definition").textValue(), parameters);
+
+    return new ConstraintDefinitionDto(name, description, defaultType, parameters, definition);
+  }
+
+  private static AbstractSyntaxTreeDto parseDefinition(String definition,
+                                                       HashMap<String, ResourceType> parameters)
+      throws de.uftos.repositories.ucdl.parser.javacc.ParseException {
+    return DefinitionParser.parseDefinition(definition, parameters);
+  }
```
danieldietzler, line 100: Why is this method needed? AFAICT it's only called once and literally only has one line of code, this doesn't even help with readability

server/ucdl/ucdl.yml
danieldietzler, line 1: Missing new line at the end of this file.

danieldietzler (commented):
Haven't looked at the `DefinitionParser` yet but everything else looks super clean actually! Much better than I expected 😅

server/src/main/java/de/uftos/dto/ucdl/ast/ControlSequenceDto.java
marcelfloe: There is no "generic AST" but the getToken- method allows to identify, which subclass is used.

server/src/main/java/de/uftos/repositories/ucdl/UcdlRepositoryImpl.java
marcelfloe: Sure

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+        SetDto set = (SetDto) buildAst(root.jjtGetChild(1), parameters);
+        variableDefinition.add(set); //set
+        OperatorDto variable = new OperatorDto(UcdlToken.OF, variableDefinition);
+
+        parameters.put(varName.value(), set.type()); //adding the new variable
+        AbstractSyntaxTreeDto body = buildAst(root.jjtGetChild(1), parameters);
+        parameters.remove(varName.value()); //removing the new variable
+
+        return new QuantifierDto(UcdlToken.EXISTS, variable, body);
+      }
```
danieldietzler, line 176: Those two are identical except for the very last line. That's generally a pattern here; that code blocks are copied multiple times. It definitely makes sense to refactor that at some point.

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+        if (root.jjtGetChild(1).jjtGetNumChildren() > 0) {
+          AbstractSyntaxTreeDto secondArgument =
+              buildAst(root.jjtGetChild(1).jjtGetChild(0), parameters);
+          List<AbstractSyntaxTreeDto> params = new ArrayList<>();
+          params.add(firstArgument);
+          params.add(secondArgument);
+          return new OperatorDto(UcdlToken.AND, params);
+        }
+        return firstArgument;
+      }
```
danieldietzler, line 212: For all operators with two params the same comment applies

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+          return new OperatorDto(UcdlToken.NOT, param);
+        }
+        return buildAst(root.jjtGetChild(0), parameters);
+      }
+      case "ISEMPTY" -> {
+        List<AbstractSyntaxTreeDto> param = new ArrayList<>();
+        param.add(buildAst(root.jjtGetChild(0), parameters));
+        return new OperatorDto(UcdlToken.IS_EMPTY, param);
+      }
+      case "ELEMENTINSETOREQUATION" -> {
```
danieldietzler, line 226: Is it allowed to have underscores in tokens? That would help with readability *a lot*, especially for `ELEMENT_IN_SET_OR_EQUATION`

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+
+        ResourceType elementType = parameters.get(elementName.value());
+
+        List<AbstractSyntaxTreeDto> attributes = new ArrayList<>();
+
+        Node attributeList = root.jjtGetChild(1);
+        while (attributeList.jjtGetNumChildren() > 0) {
+          String attribute = ((SimpleNode) attributeList).jjtGetFirstToken().next.image;
+          elementType = getNextResourceType(elementType, attribute);
+          //^ also performs semantic check whether the attribute is applicable to the given type
```
danieldietzler, line 262: Such comments are super helpful and imo they bring great value

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+                  && ((SetDto) filter).type() != setType)) {
+                throw new ParseException("Sets can only be filtered using sets of the same type!");
+              }
+            }
+
+            modifiers.add(new OperatorDto(UcdlToken.FILTER, filters));
+            modifierList = modifierList.jjtGetChild(2);
+            parameters.replace("this", thisType);
+          }
+        }
```
danieldietzler, line 332: I think this should get a massive cleanup. What about cases where `#children` is one or greater than 3? Can that never happen?

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+              filters.add(buildAst(filterList.jjtGetChild(0), parameters));
+              filterList = filterList.jjtGetChild(1);
+            }
+
+            for (AbstractSyntaxTreeDto filter : filters) {
+              if ((filter.getToken() == UcdlToken.NUMBER_SET && setType != ResourceType.NUMBER)
+                  || (filter.getToken() == UcdlToken.RESOURCE_SET
+                  && ((SetDto) filter).type() != setType)) {
+                throw new ParseException("Sets can only be filtered using sets of the same type!");
+              }
```
danieldietzler, line 325: Use early `continue`s to make it readable!
Let's apply some logic...
Let
```
A := filter.getToken() == UcdlToken.NUMBER_SET
B := setType != ResourceType.NUMBER
C := filter.getToken() == UcdlToken.RESOURCE_SET
D := ((SetDto) filter).type() != setType
```
Then the expression becomes `(A && B) || (C && D)`. Negating this leaves us with `(NOT A || NOT B) && (NOT C || NOT D) = (NOT A && NOT C) || (NOT B && NOT C) || (NOT A && NOT D) || (NOT B && NOT D)`
Thus, if either of those terms is fulfilled we want to continue the loop and not throw an error.
This way, we can have four simple, easy to understand ifs without nesting of logic or ifs.

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+        List<Integer> values = new ArrayList<>();
+
+        values.add(Integer.parseInt(((SimpleNode) root).jjtGetFirstToken().next.image));
+
+        SimpleNode numberList = (SimpleNode) root.jjtGetChild(0);
+        while (numberList.jjtGetNumChildren() > 0) {
+          values.add(Integer.parseInt(numberList.jjtGetFirstToken().next.image));
+          numberList = (SimpleNode) numberList.jjtGetChild(0);
+        }
+        return new ValueDto<>(UcdlToken.NUMBER_SET, values.toArray(new Integer[0]));
```
danieldietzler, line 346: That you can't write `values.toArray<Integer>()` still baffles me. Average java moment lol

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+                Node filterList = modifierList.jjtGetChild(1);
+                while (filterList.jjtGetNumChildren() > 0) {
+                  filters.add(buildAst(filterList.jjtGetChild(0), parameters));
+                  filterList = filterList.jjtGetChild(1);
+                }
+                for (AbstractSyntaxTreeDto filter : filters) {
+                  if (filter.getToken() == UcdlToken.RESOURCE_SET
+                      && ((SetDto) filter).type() != ResourceType.NUMBER) {
+                    throw new ParseException(
+                        "ets can only be filtered using sets of the same type!");
```
danieldietzler, line 384: `ets`?

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+          case "ELEMENT":
+            switch (root.jjtGetChild(1).toString()) {
+              case "SETMODIFICATION":
+                AbstractSyntaxTreeDto setName = buildAst(root.jjtGetChild(0), parameters);
+                ResourceType setType;
+                if (setName.getToken() == UcdlToken.NUMBER) {
+                  setType = ResourceType.NUMBER;
+                } else {
+                  setType = ((ElementDto) setName).type();
+                }
```
danieldietzler, line 401: If a ternary wouldn't become too long and line break I'd prefer that

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+                        throw new ParseException(
+                            "Sets can only be filtered using sets of the same type!");
+                      }
+                    }
+
+                    modifiers.add(new OperatorDto(UcdlToken.FILTER, filters));
+                    modifierList = modifierList.jjtGetChild(2);
+                    parameters.replace("this", thisType);
+                  }
+                }
```
danieldietzler, line 441: I *think* I've seen that code block already, but it's also just very complex

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+            }
+          default:
+            throw new IllegalStateException();
+
+        }
+      }
+      //includes cases: "CONTROLSEQUENCELIST", "CONTROLSEQUENCERETURN",
+      // "OPTIONALIMPLIES", "OPTIONALOR", "OPTIONALAND",
+      // "ELEMENTINSET", "EQUATION", "ELEMENTEQUATION", "ELEMENTATTRIBUTELIST",
+      // "SETNAME", "SETMODIFICATION", "NUMBERLIST", "ATTRIBUTE", "FILTERLIST"
```
danieldietzler, line 493: You don't mean the default case includes those cases, do you?

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+      default -> throw new IllegalStateException();
+    }
+  }
+
+  //checks that all returnValues are the same and returns the value
+  // (throws an exception if true and false are returnValues inside the body)
```
danieldietzler, line 499: When you already want to comment it a proper javadoc might be nicer

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+  private static boolean getReturnValue(List<AbstractSyntaxTreeDto> body) throws ParseException {
+    boolean and = true;
+    boolean or = false;
+    for (AbstractSyntaxTreeDto ast : body) {
+      if (ast.getToken() == UcdlToken.FOR || ast.getToken() == UcdlToken.IF) {
+        ControlSequenceDto cs = (ControlSequenceDto) ast;
+        and = and && cs.returnValue();
+        or = or || cs.returnValue();
```
danieldietzler, line 507:
```suggestion
        and &= cs.returnValue();
        or |= cs.returnValue();
```

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+      }
+    }
+    if (and != or) {
```
danieldietzler, line 510: That is clever actually

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+  }
+
+  private static UcdlToken getComparatorToken(String comparator, List<AbstractSyntaxTreeDto> params)
+      throws ParseException {
+    UcdlToken token = switch (comparator) {
+      case ">" -> UcdlToken.GREATER;
+      case "<" -> UcdlToken.SMALLER;
+      case ">=" -> UcdlToken.GREATER_EQUALS;
+      case "<=" -> UcdlToken.SMALLER_EQUALS;
+      case "=", "==" -> UcdlToken.EQUALS;
```
danieldietzler, line 525: Shouldn't we reserve `=` in case ucdl should ever get assignments? Also, a single equal sign is just confusing to me 😅

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+      case "!=" -> UcdlToken.NOT_EQUALS;
+      default -> throw new IllegalStateException();
+    };
+    if (token == UcdlToken.GREATER || token == UcdlToken.SMALLER
+        || token == UcdlToken.GREATER_EQUALS || token == UcdlToken.SMALLER_EQUALS) {
+      for (AbstractSyntaxTreeDto ast : params) {
+
```
danieldietzler, line 532:
```suggestion
```

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+        if (ast.getToken() != UcdlToken.NUMBER && ast.getToken() != UcdlToken.SIZE
+            && !(ast.getToken() == UcdlToken.ELEMENT
+            && ((ElementDto) ast).type() == ResourceType.NUMBER)) {
```
danieldietzler, line 535: Shouldn't this be "if the type is *not* `ResourceType.NUMBER`"?

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+                  + "\"<\",\">\",\"<=\",\">=\"!");
+        }
+      }
+    }
+    return token;
+  }
+
+  private static ResourceType getNextResourceType(ResourceType currentType, String attribute)
+      throws ParseException {
+    return switch (attribute) {
```
danieldietzler, line 547: It would be awesome if we could get that switch from the actual code. Duplicating the whole structure here is kinda annoying imo. *However*, I'm not sure if it's worth it lol

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+              "Attribute \"" + attribute
+                  + "\" is only available for \"this\"(outside filters), students, teachers, "
+                  + "studentGroups, rooms, subjects grades and timeslots!");
+        }
+        yield ResourceType.TAG;
+      }
+      default -> throw new ParseException("Illegal attribute name: " + attribute);
+    };
+  }
+}
```
danieldietzler, line 691: Missing new line at the end

danieldietzler (commented):
TBH I didn't completely understand (and thus validate) the logic of the parser. I tried my best and I *think* it looks good, but it's also very complex, so.... My focus was on readability and refactoring stuff for the most part

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
marcelfloe: It could be part of Syntax-Checking with a different grammar, but I would leave it as is.
After all it works this way.

server/src/main/java/de/uftos/repositories/ucdl/UcdlRepositoryImpl.java
```diff
@@ -11,24 +16,41 @@
...
 @Repository
 public class UcdlRepositoryImpl implements UcdlRepository {
+  private static final Path UCDL_PATH = Paths.get("/app/ucdl/ucdl.yml");
+  private HashMap<String, ConstraintDefinitionDto> currentDefinitions = null;
 
   @Override
-  public String getUcdl() {
-    return "";
+  public String getUcdl() throws IOException {
+    Files.createFile(UCDL_PATH);
```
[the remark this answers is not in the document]
marcelfloe: "Creates a new and empty file, failing if the file already exists. The check for the existence of the file and the creation of the new file if it does not exist are a single operation that is atomic with respect to all other filesystem activities that might affect the directory."

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+        List<AbstractSyntaxTreeDto> body = new ArrayList<>();
+        body.add(buildAst(root.jjtGetChild(2), parameters)); //first control sequence
+
+        Node controlSequenceList = root.jjtGetChild(3); //rest of the control sequences
+        while (controlSequenceList.jjtGetNumChildren() == 2) {
+          body.add(buildAst(controlSequenceList.jjtGetChild(0), parameters));
+          controlSequenceList = controlSequenceList.jjtGetChild(1);
+        }
+
+        parameters.remove(varName.value()); //removing the new variable
```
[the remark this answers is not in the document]
marcelfloe: Technically you could use the variable outside of its namespace if it isn`t removed.
For example:

for (lesson : this.lessons) {code}
if (lesson.slot < 3) {code}
return bool

would be valid code (at least as far as the semantics - check can tell.

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -0,0 +1,691 @@
...
+
+        //semantic check for legal return values
+        boolean returnValue = getReturnValue(body);
+        return new ControlSequenceDto(UcdlToken.FOR, variable, body, returnValue);
+      }
+      case "IF" -> {
+        boolean returnValue = false;
```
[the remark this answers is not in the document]
marcelfloe: Do you mean the variable "returnValue"?

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
marcelfloe: Yes, '_' is allowed. This will require generating a new parser.

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
marcelfloe: I guess that's sarcasm. But I can't tell for sure...

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
marcelfloe: The amount of children is directly tied to the grammar.
So no, if the AST given was parsed using the javacc-parser than it is not possible to have different a differen amount of children than 0, 2 or 3. 
(The specification in the grammar file: 
void SETMODIFICATION(): {} { <FLATMAP> ATTRIBUTE() SETMODIFICATION() | <LBRACKET> FILTER() FILTERLIST() <RBRACKET> SETMODIFICATION() | {} }
)

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
marcelfloe: You can also just use  `values.toArray()`, however this would just return an `Object[]`.

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
marcelfloe: I fixed this typo once before...
I guess it was a different branch...

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
marcelfloe: Probably also exists in `case "SET"`...
I need to refactor a lot of code duplication.

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
marcelfloe: It does.
Those are part of other cases and do not align with the general AST structure and as the method is required to return a valid AST it would not make sense to allow for those cases to be executed directly.
However they are still listed to make sure that every possible case is taken care of.

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
marcelfloe: Didn't know these operators existed.

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
marcelfloe: Thanks

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
marcelfloe: I think it was left in for users who might not be familiar with programming.
But we can remove it (however this will require generating a new parser).

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
marcelfloe: It is already negated: `!(ast.getToken() == UcdlToken.ELEMENT && ((ElementDto) ast).type() == ResourceType.NUMBER)`

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
marcelfloe: In fact: it was the solver branch...

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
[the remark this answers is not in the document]
marcelfloe: I don't think I understand what you mean exactly...

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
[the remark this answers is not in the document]
marcelfloe: Unless we clone the HashMap whenever a parameter is added, but I don`t think, that it would perform better...

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
danieldietzler: Ah, so basically those are valid tokens, but we shouldn't run into them here. And if we do, something went completely wrong. Thanks for the explanation!

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
danieldietzler: Yes

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
danieldietzler: Oh, I overlooked that that whole expression is in parenthesis. This is probably another place where you should refactor the condition :P

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
danieldietzler: Got it

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
danieldietzler: Exactly. I'm annoyed that you can't just type it using generics

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
danieldietzler: No, I was actually serious lol. That's one of a few comments I actually really like :)

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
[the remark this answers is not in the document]
danieldietzler: In the `RETURN` case you can `return new ControlSequenceDto(UcdlToken.IF, bool, body, dto.value())` an in the `CONTROLSEQUENCE` case you'd `return new ControlSequenceDto(UcdlToken.IF, bool, body, getReturnValue(body))`

server/src/main/java/de/uftos/repositories/ucdl/UcdlRepositoryImpl.java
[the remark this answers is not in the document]
danieldietzler: So this should be wrapped in a try catch then, correct? As otherwise that exception will just be thrown up all the way which we definitely don't want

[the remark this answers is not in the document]
marcelfloe: The question would be what we want to do, when we catch the exception. (In general for all IOExceptions thrown when trying to interact with ucdl.yml)

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
marcelfloe: Same.
Generics would be cleaner.

server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
marcelfloe: It would.

danieldietzler (approved):
Thank you very much for putting all this effort into this!

PR #147, opened 2024-07-15
Feat/curriculum service todos

jdiester:


danieldietzler (approved):
LGTM

server/src/main/java/de/uftos/controller/ConstraintController.java
```diff
@@ -27,6 +29,8 @@
 public class ConstraintController {
   private final ConstraintSignatureService constraintSignatureService;
   private final ConstraintInstanceService constraintInstanceService;
+  private final ConstraintInstanceRepository constraintInstanceRepository;
+  private final ConstraintSignatureRepository constraintSignatureRepository;
```
SuperMarcomen, line 33: Why is this needed?

server/src/main/java/de/uftos/dto/CurriculumRequestDto.java
```diff
@@ -11,15 +12,18 @@
...
   /**
    * Maps the information from the data transfer object to a new curriculum entity.
    *
+   * @param grade the Grade object to be mapped
    * @return the new curriculum entity.
    */
-  public Curriculum map() {
-    return new Curriculum(this.gradeId, this.lessonsCounts);
+  public Curriculum map(Grade grade) {
+    return new Curriculum(grade, this.name, this.lessonsCounts);
```
SuperMarcomen, line 26: I think that you don't need the Grade instance and that
```java
return new Curriculum(new Grade(gradeId), this.name, this.lessonsCounts);
```


should work just fine

server/src/main/java/de/uftos/dto/CurriculumResponseDto.java
```diff
@@ -16,4 +18,19 @@
...
+
+
+  /**
+   * Creates a CurriculumResponseDto from a Curriculum.
+   *
+   * @param curriculum the Curriculum whose information should be included
+   *                   in the CurriculumResponseDto.
+   * @return the CurriculumResponseDto containing information about the Curriculum.
+   */
+  public static CurriculumResponseDto createResponseDtoFromCurriculum(Curriculum curriculum) {
```
SuperMarcomen, line 30: Shouldn't this be a constructor?

SuperMarcomen (changes requested):

SuperMarcomen (approved):

PR #158, opened 2024-07-18
feat: constraint endpoint

SuperMarcomen:
…nstraints/{signatureId}/instances

### DIFF
--- a/server/src/main/java/de/uftos/services/ConstraintInstanceService.java
+++ b/server/src/main/java/de/uftos/services/ConstraintInstanceService.java
@@ -1,11 +1,15 @@
 package de.uftos.services;
 
+import de.uftos.controller.ConstraintArgumentDisplayName;
 import de.uftos.dto.ConstraintInstanceRequestDto;
+import de.uftos.dto.ConstraintInstanceResponseDto;
 import de.uftos.dto.ResourceType;
 import de.uftos.entities.ConstraintArgument;
 import de.uftos.entities.ConstraintInstance;
 import de.uftos.entities.ConstraintParameter;
 import de.uftos.entities.ConstraintSignature;
+import de.uftos.entities.Student;
+import de.uftos.entities.Teacher;
 import de.uftos.repositories.database.ConstraintInstanceRepository;
 import de.uftos.repositories.database.ConstraintSignatureRepository;
 import de.uftos.repositories.database.GradeRepository;
@@ -19,6 +23,7 @@ import de.uftos.repositories.database.TeacherRepository;
 import de.uftos.repositories.database.TimeslotRepository;
 import de.uftos.utils.SpecificationBuilder;
 import java.util.AbstractMap;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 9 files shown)


server/src/main/java/de/uftos/services/ConstraintInstanceService.java
```diff
@@ -123,16 +128,21 @@ public ConstraintInstance create(String signatureId, ConstraintInstanceRequestDt
...
+    //  GOOD LUCK!
     Specification<ConstraintInstance> specification = new SpecificationBuilder<ConstraintInstance>()
-        .andJoinEquals(signatureId, "signature", "name")
-        .optionalAndJoinLikeIgnoreCase(argument, "arguments", "value")
         .build();
 
-    return this.repository.findAll(specification, pageable);
+    Page<ConstraintInstance> constraintInstances = this.repository.findAll(specification, pageable);
+    List<ConstraintArgumentDisplayName> displayNames =
+        processConstraintInstances(constraintInstances.getContent());
```
danieldietzler, line 144: `process` is like extremely generic. Could we be more specific in the naming here?

server/src/main/java/de/uftos/services/ConstraintInstanceService.java
```diff
@@ -177,11 +186,67 @@ public ConstraintInstance update(String signatureId, String id,
...
+        Teacher teacher = teacherRepository.findById(id).orElseThrow();
+        yield new ConstraintArgumentDisplayName(id,
+            "%s %s".formatted(teacher.getFirstName(), teacher.getLastName()));
+      }
+      case STUDENT_GROUP -> new ConstraintArgumentDisplayName(id,
+          studentGroupRepository.findById(id).orElseThrow().getName());
+      case TIMESLOT -> new ConstraintArgumentDisplayName(id,
+          timeslotRepository.findById(id).orElseThrow().getDay().toString());
+    };
+  }
```
danieldietzler, line 249: Gotta love those lol

server/src/main/java/de/uftos/services/ConstraintInstanceService.java
```diff
@@ -177,11 +186,67 @@ public ConstraintInstance update(String signatureId, String id,
...
+    return displayNames;
+  }
+
+  private ConstraintArgumentDisplayName getDisplayName(ConstraintArgument constraintArgument) {
+    String id = constraintArgument.getValue();
+    return switch (constraintArgument.getConstraintParameter().getParameterType()) {
+      case TAG ->
+          new ConstraintArgumentDisplayName(id, tagRepository.findById(id).orElseThrow().getName());
+      case ROOM -> new ConstraintArgumentDisplayName(id,
+          roomRepository.findById(id).orElseThrow().getName());
```
danieldietzler, line 227: We shouldn't have empty `orElseThrow` calls. Either remove them, use `.get()` and accept the runtime error (which I'm not sure if that's fine or not), or throw a bad request.

server/src/main/java/de/uftos/services/ConstraintInstanceService.java
SuperMarcomen: done

server/src/main/java/de/uftos/services/ConstraintInstanceService.java
SuperMarcomen: Now throwing a bad request

server/src/main/java/de/uftos/services/ConstraintInstanceService.java
SuperMarcomen: I tried to make the switch as tidy and simple as possible. I don't know if it's possible to improve it

danieldietzler: Yeah I don't think so either tbh

danieldietzler (approved):
Ty!

PR #161, opened 2024-07-19
Parser Update

marcelfloe:
Added semantics check for "==" and "!=".
Fixed/added semantics check for "element in set".
Fixed quantifier.

### DIFF
--- a/server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
+++ b/server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
@@ -159,28 +159,14 @@ public class DefinitionParser {
       case "ELEMENT_IN_SET_OR_EQUATION" -> {
         List<AbstractSyntaxTreeDto> params = new ArrayList<>();
         AbstractSyntaxTreeDto element = buildAst(root.jjtGetChild(0), parameters);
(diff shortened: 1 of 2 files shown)


server/src/main/java/de/uftos/repositories/ucdl/parser/DefinitionParser.java
```diff
@@ -304,14 +327,33 @@ private static AbstractSyntaxTreeDto buildAst(Node root, HashMap<String, Resourc
...
                 List<AbstractSyntaxTreeDto> params = new ArrayList<>();
-                ElementDto element = (ElementDto) buildAst(root.jjtGetChild(0), parameters);
+                AbstractSyntaxTreeDto element = buildAst(root.jjtGetChild(0), parameters);
+                ResourceType elementType;
+                if (element.getToken() == UcdlToken.NUMBER
+                    || element.getToken() == UcdlToken.SIZE) {
+                  elementType = ResourceType.NUMBER;
+                } else {
+                  elementType = ((ElementDto) element).type();
+                }
```
danieldietzler, line 337: This and the if below for sets are used multiple times across the code. I think it makes sense to introduce a `getType` and `getSetType` private function in here

danieldietzler (approved):
LGTM

PR #162, opened 2024-07-20
chore: fix checkstyle

danieldietzler:


marcelfloe (approved):
I will enjoy the new limit of 120 characters instead of 100 characters per line.

PR #168, opened 2024-07-21
Feat: curriculum page

Lama64:


web/src/routes/admin/(resources)/curriculums/+page.ts
```diff
@@ -1,9 +1,53 @@
...
+      const curriculumRequestDto: CurriculumRequestDto = {
+        gradeId: grade.id,
+        lessonsCounts: subjects.map((subject) => ({ count: 0, subjectId: subject.id })),
+        name: grade.name,
+      };
+      const result = await createCurriculum(curriculumRequestDto);
+      grade.curriculumId = result.id;
+    }
+  }
+  grades = await getGrades({ sort: ['name,asc'] });
```
danieldietzler, line 45: Why do you fetch the grades again here? Also, you can use your `sort` object here as well, if that line is necessary.

danieldietzler (approved):
Awesome, just one small question :)

PR #171, opened 2024-07-21
feat: overview page

SuperMarcomen:


web/src/lib/components/ui/card/card.svelte
```diff
@@ -0,0 +1,95 @@
...
+            fill="white"
+            stroke="#2B6E89"
+            stroke-width="0.25"
+          />
+        </svg>
```
danieldietzler, line 46: We shouldn't paste the raw SVG but instead just use an existing icon library. Where do you have those icons from?

web/src/lib/elements/ui/link/link.svelte
```diff
@@ -0,0 +1,22 @@
...
+  import { onMount } from 'svelte';
+  import { Button } from '../button';
+
+  export let text: string;
+  export let url: string;
+  let active: boolean = false;
+
+  onMount(() => {
+    active = window.location.pathname === url;
+  });
```
danieldietzler, line 11: You can get rid of this and simply use `$: active = $page.url === url`

web/src/lib/elements/ui/link/link.svelte
```diff
@@ -0,0 +1,22 @@
...
+</script>
+
+<Button
+  on:click={() => {
+    window.location.href = url;
```
danieldietzler, line 16: You should use `goto(url)` here

web/src/lib/elements/ui/link/link.svelte
```diff
@@ -0,0 +1,22 @@
...
+  export let text: string;
+  export let url: string;
+  let active: boolean = false;
+
+  onMount(() => {
+    active = window.location.pathname === url;
+  });
+</script>
+
+<Button
```
danieldietzler, line 14: We shouldn't be using a button for an link, that's bad for ARIA. Instead just use an `<a>` tag

web/src/routes/admin/+page.svelte
```diff
@@ -1,2 +1,118 @@
...
+
+  onMount(async () => {
+    pageLoaded = true;
+    stats = await loadStats();
+  });
+
+  async function loadStats() {
+    const response = await getServerStats();
+    return response;
+  }
```
danieldietzler, line 40: You should wrap the affected html (the part that requires the response) in `{# await getServerStats() then stats}` and work with that.

web/src/lib/utils.ts
```diff
@@ -25,6 +25,24 @@ export const WorkingHours = [
...
 
+export enum Links {
+  STUDENT,
+  TEACHER,
+  CLASS,
+  ROOM,
+  STUDENT_GROUP,
+  GRADE,
+  CURRICULUM,
+}
```
danieldietzler, line 36: Maybe I am blind but where's this enum used?

web/src/lib/components/ui/card/card.svelte
SuperMarcomen: I got them from a figma plugin. I don't think we can find them in any library.

web/src/lib/elements/ui/link/link.svelte
SuperMarcomen: done

web/src/routes/admin/+page.svelte
SuperMarcomen: thank you, done

web/src/lib/utils.ts
SuperMarcomen: You're not blind, it is no longer needed and I forgot to remove it. Thank you

web/src/lib/components/ui/card/card.svelte
SuperMarcomen: I found some alternatives. Not the best, but they work.

danieldietzler (approved):
LGTM

PR #181, opened 2024-07-23
String parser

marcelfloe:
Including a method to parse any String.


danieldietzler (approved):
Awesome, thank you very much!

PR #191, opened 2024-07-24
Feat/timeslot tests

jdiester:


server/src/main/java/de/uftos/controller/TimeslotController.java
```diff
@@ -47,13 +49,14 @@ public Timeslot createTimeslot(@RequestBody TimeslotRequestDto timeslot) {
...
+   * {@link TimeslotService#get(Pageable, Optional) get} function of the timeslot service.
    *
-   * @return all timeslots.
+   * @param pageable contains the parameters for the page.
+   * @return the page of timeslots fitting the parameters.
    */
   @GetMapping()
-  public List<Timeslot> getTimeslots() {
-    return this.timeslotsService.get();
+  public Page<Timeslot> getTimeslots(Pageable pageable, Optional<String[]> tags) {
```
danieldietzler, line 58: I don't think we need paginated timeslots do we?

danieldietzler (approved):
LGTM, thank you :)

PR #199, opened 2024-07-26
(Hopefully) finished solver

### LINKED ISSUE #45: server: solver adapter

marcelfloe:
closes #45


server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
```diff
@@ -0,0 +1,1073 @@
...
+    return (parameters) -> {
+      ResourceTimefoldInstance first = firstElement.apply(parameters);
+      ResourceTimefoldInstance second = secondElement.apply(parameters);
+
+      if (first.getResourceType() != ResourceType.NUMBER
+          || second.getResourceType() != ResourceType.NUMBER) {
+        throw new IllegalArgumentException();
+      }
+
+      return ((Number) first).getValue() < ((Number) second).getValue();
```
danieldietzler, line 514: Here the same question

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
```diff
@@ -0,0 +1,1073 @@
...
+            convertSet(param, params);
+        Function<List<ResourceTimefoldInstance>, ResourceTimefoldInstance> element =
+            convertValueReference(new ValueDto<>(UcdlToken.VALUE_REFERENCE, "this"), params);
+        functions.add((parameters -> set.apply(parameters).contains(element.apply(parameters))));
+      } else {
+        functions.add(convertBool(param, params));
+      }
+    }
+
+    return (parameters -> set -> {
```
danieldietzler, line 628: What is happening there? That's a function taking parameters and mapping to another function that takes `set` as an argument?

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
```diff
@@ -0,0 +1,1073 @@
...
+      ResourceTimefoldInstance thisPointer = parameters.getFirst();
+      Set<ResourceTimefoldInstance> filteredSet = new HashSet<>();
+      outerLoop:
+      for (ResourceTimefoldInstance resource : set) {
+        parameters.set(0, resource); //replacing "this"-pointer
+        for (Function<List<ResourceTimefoldInstance>, Boolean> function : functions) {
+          if (!function.apply(parameters)) {
+            continue outerLoop;
```
danieldietzler, line 636: Uuuuhm. Can't you just`break` out of the inner loop? Goto is honestly extremely ugly and prone to confusion

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
```diff
@@ -0,0 +1,1073 @@
...
+    if (set.setName().getToken() == UcdlToken.VALUE_REFERENCE) {
+      name = convertValueReference(set.setName(), params);
+    } else {
+      name = convertElement(set.setName(), params);
+    }
+
+    List<Function<
+        List<ResourceTimefoldInstance>,
+        Function<Set<ResourceTimefoldInstance>, Set<ResourceTimefoldInstance>>
+        >> modifiers = new ArrayList<>();
```
danieldietzler, line 703: This formatting is very much screwed up. Can you somehow make it better readable?

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
```diff
@@ -0,0 +1,1073 @@
...
+    }
+
+    return (parameters) -> {
+      Set<ResourceTimefoldInstance> setReference = new HashSet<>();
+      setReference.add(name.apply(parameters));
+
+      for (Function<
+            List<ResourceTimefoldInstance>,
+            Function<Set<ResourceTimefoldInstance>, Set<ResourceTimefoldInstance>>
+            > function : modifiers) {
```
danieldietzler, line 720: Honestly, you should just use `var` here. Java doesn't allow you to specify type aliases because it's stupid so I think `var` is the best solution

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
```diff
@@ -0,0 +1,1073 @@
...
+  private static Function<List<ResourceTimefoldInstance>, Set<ResourceTimefoldInstance>>
+      convertNumberSet(AbstractSyntaxTreeDto ast, LinkedHashMap<String, ResourceType> params) {
+    if (ast.getToken() != UcdlToken.NUMBER_SET) {
+      throw new IllegalStateException();
+    }
+    ValueDto<Integer[]> set = (ValueDto<Integer[]>) ast;
+
+    Set<ResourceTimefoldInstance> resources = new HashSet<>();
+
+    for (Integer i : set.value()) {
```
danieldietzler, line 736:
```suggestion
    for (int i : set.value()) {
```
You shouldn't use wrapper types if avoidable

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
```diff
@@ -0,0 +1,1073 @@
...
+      case "index" -> (parameters) -> (set) -> {
+        Set<ResourceTimefoldInstance> resultSet = new HashSet<>();
+        for (ResourceTimefoldInstance resource : set) {
+          if (resource.getResourceType() == ResourceType.LESSON) {
+            resultSet.add(new Number(((LessonTimefoldInstance) resource).getIndex()));
+          } else {
+            throw new IllegalStateException();
+          }
+        }
+        return resultSet;
```
danieldietzler, line 764: Might be worth moving all of those into their own functions as well

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
```diff
@@ -0,0 +1,1073 @@
...
+      };
+      case "students" -> (parameters) -> (set) -> {
+        Set<ResourceTimefoldInstance> resultSet = new HashSet<>();
+        for (ResourceTimefoldInstance resource : set) {
+          switch (resource.getResourceType()) {
+            case TIMETABLE ->
+                resultSet.addAll(((TimetableSolutionTimefoldInstance) resource).getStudents());
```
danieldietzler, line 859: `resource` is a single item, right? Why do you need `addAll` then? AM I overlooking something?

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
```diff
@@ -0,0 +1,1073 @@
...
+    }
+
+    return (parameters) -> {
+      Set<ResourceTimefoldInstance> elementReference = new HashSet<>();
+      elementReference.add(name.apply(parameters));
+
+      for (Function<
+            List<ResourceTimefoldInstance>,
+            Function<Set<ResourceTimefoldInstance>, Set<ResourceTimefoldInstance>>
+            > function : attributes) {
```
danieldietzler, line 1023: Those types are killing me lmao

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
```diff
@@ -0,0 +1,1073 @@
...
+  private static Function<List<ResourceTimefoldInstance>, Number> convertNumberElement(
+      AbstractSyntaxTreeDto ast, LinkedHashMap<String, ResourceType> params) {
+    return switch (ast.getToken()) {
+      case NUMBER -> convertNumber(ast, params);
+      case SIZE -> convertSize(ast, params);
+      default -> throw new IllegalStateException();
+    };
+  }
+
+  private static Function<List<ResourceTimefoldInstance>, Number> convertNumber(
```
danieldietzler, line 1042: I do quite like these `convertX` methods. They are very well contained and much better readable than those huuuuuge single methods.

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionTimefoldInstance.java
```diff
@@ -0,0 +1,20 @@
...
+ * @param name               the name of the constraint definition.
+ * @param defaultType        the default hard/soft penalize/reward value for this constraint definition.
+ * @param evaluationFunction the function, which is used to evaluate this constraint.
+ */
+public record ConstraintDefinitionTimefoldInstance(
+    String name,
+    RewardPenalize defaultType,
+    Function<List<ResourceTimefoldInstance>, Boolean> evaluationFunction
+) {
+}
```
danieldietzler, line 20: Missing new line

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/constraintinstances/ConstraintInstanceHardPenalize.java
```diff
@@ -0,0 +1,38 @@
...
+  public boolean evaluate(HashMap<String, ResourceTimefoldInstance> resources) {
+    List<ResourceTimefoldInstance> params = new ArrayList<>();
+    for (ResourceTimefoldInstance resource : parameters) {
+      params.add(resources.get(resource.getId()));
+    }
+    try {
+      return evaluationFunction.apply(params);
+    } catch (NullPointerException e) {
+      return true;
+    }
```
danieldietzler, line 31: Catching a null pointer exception here look sketchy. Why do you need that? You should much rather just do `if (something == null) { return true; }`

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/constraintinstances/ConstraintInstanceSoftReward.java
```diff
@@ -0,0 +1,38 @@
...
+  @Override
+  public boolean evaluate(HashMap<String, ResourceTimefoldInstance> resources) {
+    List<ResourceTimefoldInstance> params = new ArrayList<>();
+    for (ResourceTimefoldInstance resource : parameters) {
+      params.add(resources.get(resource.getId()));
+    }
+    try {
+      return evaluationFunction.apply(params);
+    } catch (NullPointerException e) {
+      return false;
```
danieldietzler, line 30: Same for all the other records

server/src/main/java/de/uftos/repositories/solver/timefold/domain/LessonTimefoldInstance.java
```diff
@@ -0,0 +1,92 @@
...
+  @JsonIdentityReference
+  private SubjectTimefoldInstance subject;
+  @JsonIdentityReference
+  @PlanningVariable
+  private RoomTimefoldInstance room;
+
+  //required by the timefold solver
+  public LessonTimefoldInstance() {
+
+  }
```
danieldietzler, line 39: Replace this with lombok's `@NoArgsConstructor`

server/src/main/java/de/uftos/repositories/solver/timefold/domain/LessonTimefoldInstance.java
```diff
@@ -0,0 +1,92 @@
...
+  @JsonIdentityReference
+  @PlanningVariable
+  private TeacherTimefoldInstance teacher;
+  @JsonIdentityReference
+  private StudentGroupTimefoldInstance studentGroup;
+  @JsonIdentityReference
+  private SubjectTimefoldInstance subject;
+  @JsonIdentityReference
+  @PlanningVariable
+  private RoomTimefoldInstance room;
```
danieldietzler, line 34: For one thing, new lines between the fields would help. Also, what is this json magic doing?

server/src/main/java/de/uftos/repositories/solver/timefold/domain/TimetableSolutionTimefoldInstance.java
```diff
@@ -0,0 +1,96 @@
...
+  @ProblemFactCollectionProperty
+  @ValueRangeProvider
+  private final List<TagTimefoldInstance> tags = new ArrayList<>();
+  @PlanningEntityCollectionProperty
+  private final List<LessonTimefoldInstance> lessons = new ArrayList<>();
+  @ProblemFactCollectionProperty
+  @ValueRangeProvider
+  private final List<ConstraintInstanceTimefoldInstance> constraintInstances = new ArrayList<>();
+  @PlanningScore
+  private HardSoftScore score;
```
danieldietzler, line 53: New lineeeeeeees

server/src/main/java/de/uftos/repositories/ucdl/parser/PredefinedConstraint.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.List;
+import lombok.Getter;
+
+
+/**
+ * This enum models predefined constraints which are ready-to-use.
+ */
+@SuppressWarnings("checkstyle:LineLength")
+@Getter
+public enum PredefinedConstraint {
```
danieldietzler, line 14: I am not sure if an enum is the best type here. Also, what do you think about reading the default constraints from a file?

danieldietzler (changes requested):
Oof.

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
```diff
@@ -1,25 +1,518 @@
...
+        case STUDENT ->
+            students.put(resource.getId(), new StudentTimefoldInstance(resource.getId()));
+        case SUBJECT ->
+            subjects.put(resource.getId(), new SubjectTimefoldInstance(resource.getId()));
+        case TAG -> tags.put(resource.getId(), new TagTimefoldInstance(resource.getId()));
+        case TEACHER ->
+            teachers.put(resource.getId(), new TeacherTimefoldInstance(resource.getId()));
+        case TIMESLOT ->
+            timeslots.put(resource.getId(), new TimeslotTimefoldInstance(resource.getId()));
+        default -> throw new IllegalStateException();
```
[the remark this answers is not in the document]
marcelfloe: I think i got a warning for a missing default-statement, but i'm not sure.
I will take a look at it.

.idea/.gitignore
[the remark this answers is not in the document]
marcelfloe: I have no idea, how that did end up there...

server/src/main/java/de/uftos/dto/solver/GradeProblemDto.java
```diff
@@ -11,6 +11,11 @@ public record GradeProblemDto(
     List<String> tagIds,
     List<String> studentGroupIds
 ) implements ResourceProblemDto {
+  @Override
+  public String getId() {
+    return id;
+  }
```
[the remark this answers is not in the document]
marcelfloe: Those are required for the ResourceProblemDto-interface.

server/src/main/java/de/uftos/dto/ucdl/ast/OperatorDto.java
```diff
@@ -12,4 +12,20 @@ public record OperatorDto(UcdlToken token, List<AbstractSyntaxTreeDto> parameter
...
+
+    for (AbstractSyntaxTreeDto ast : this.parameters) {
+      stringBuilder.append(ast.toString());
+      stringBuilder.append("\n");
+    }
+
+    stringBuilder.append("}");
+
+    return stringBuilder.toString();
+  }
```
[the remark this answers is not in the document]
marcelfloe: These toString-methods actually are left overs, which i forgot to remove.

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
```diff
@@ -1,25 +1,518 @@
...
  */
-@Repository
 public class SolverRepositoryImpl implements SolverRepository {
-  public Future<TimetableSolutionDto> solve(
-      TimetableProblemDto timetable,
-      HashMap<String, ConstraintDefinitionDto> constraintDefinitions,
-      List<ConstraintInstanceDto> constraintInstances
-  )
+
+  @SuppressWarnings("checkstyle:VariableDeclarationUsageDistance")
```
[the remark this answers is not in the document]
marcelfloe: Its a warning about the distance between a Variable-Declaration and its first usage.
In this case many HashMaps are declared together at the start of the method and filled below one after the other and the lower HashMaps have a distance between declaration and filling which is to big.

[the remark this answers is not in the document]
marcelfloe: I think it is actually only to big for the constraintInstances-List.

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
```diff
@@ -1,25 +1,518 @@
...
+        case TAG -> tags.put(resource.getId(), new TagTimefoldInstance(resource.getId()));
+        case TEACHER ->
+            teachers.put(resource.getId(), new TeacherTimefoldInstance(resource.getId()));
+        case TIMESLOT ->
+            timeslots.put(resource.getId(), new TimeslotTimefoldInstance(resource.getId()));
+        default -> throw new IllegalStateException();
+      }
+    }
+
+    //connecting all resources as defined by the Dtos (and checking connection)
```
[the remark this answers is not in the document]
marcelfloe: As the loop above filled the HashMaps with resources, which only have an ID the loops below manage the relations between the resources and check for consistency.

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
```diff
@@ -1,25 +1,518 @@
...
+
+    for (GradeProblemDto grade : timetable.grades()) {
+      GradeTimefoldInstance timefoldInstance = grades.get(grade.id());
+      for (String studentGroupId : grade.studentGroupIds()) {
+        StudentGroupTimefoldInstance studentGroup = studentGroups.get(studentGroupId);
+        if (studentGroup.getGrade() != null) {
+          throw new IllegalArgumentException();
+        }
```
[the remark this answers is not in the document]
marcelfloe: In case of data-inconsistencies the grade might reference a student group which doesn't exist.
In this case the HashMap would return null.
If the data is consistent this IllegalArgumentException should not be thrown.

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
```diff
@@ -1,25 +1,518 @@
...
+        tag.getTimeslotList().add(timefoldInstance);
+        timefoldInstance.getProvidedTagsList().add(tag);
+      }
+    }
+    solution.getTimeslots().addAll(timeslots.values());
+    resources.putAll(timeslots);
+
+    for (LessonProblemDto lesson : timetable.lessons()) {
+      LessonTimefoldInstance timefoldInstance = lessons.get(lesson.id());
+      timefoldInstance.setIndex(lesson.index());
```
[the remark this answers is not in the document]
marcelfloe: The code above or below?

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
```diff
@@ -0,0 +1,1073 @@
...
+      function = convertCodeblock(ast, params);
+    } else {
+      function = convertBool(ast, params);
+    }
+
+    return new ConstraintDefinitionTimefoldInstance(name, defaultType, function);
+  }
+
+  //The first parameter "List<ResourceTimefoldInstance" contains always the parameters given in the
+  //constraint definition.
```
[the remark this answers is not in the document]
marcelfloe: No, the entire List.
Each item in the List is one argument which is given to the evaluation function for evaluation.
So the (first) List given to the function which is returned by the methods below is the list of arguments, which is provided by the solver (-> ResourceTimefoldInstances)
There once where methods `convertXYZ()`which had multiple Lists in the signature of the returned function, but i think those are all sets by now.

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
```diff
@@ -0,0 +1,1073 @@
...
+
+    Function<List<ResourceTimefoldInstance>, Boolean> function = convertBool(root.body(), params);
+
+    params.remove(variableReference.value());
+
+    return (parameters) -> {
+      boolean and = true;
+      for (ResourceTimefoldInstance resource : variableDefinition.apply(parameters)) {
+        parameters.add(resource);
+        and = and && function.apply(parameters);
```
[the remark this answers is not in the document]
marcelfloe: Yes this is intentional.
This is only called in constraint evaluation and therefore all these returned functions are not allowed to have any side-effects.
So it technically doen`t matter whether the second part is called or not (but this way round it should be faster).

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
```diff
@@ -0,0 +1,1073 @@
...
+    return (parameters) -> {
+      ResourceTimefoldInstance first = firstElement.apply(parameters);
+      ResourceTimefoldInstance second = secondElement.apply(parameters);
+
+      if (first.getResourceType() != ResourceType.NUMBER
+          || second.getResourceType() != ResourceType.NUMBER) {
+        throw new IllegalArgumentException();
+      }
+
+      return ((Number) first).getValue() <= ((Number) second).getValue();
```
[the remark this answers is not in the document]
marcelfloe: There is no `.getValue()` for ResourceTimefoldInstances.
Also this number is a new implementation, which implements the ResourceTimefoldInstance-interface, but does not implement any number-interfaces (so a different class-name would probably help).

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
marcelfloe: Same answer.

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
marcelfloe: Basically it is a method which returns a function which returns a function.
It could also be seen as a method which returns a function, that takes 2 parameters.

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
marcelfloe: Only breaking the inner loop will still result in `filteredSet.add(resource);` being executed which it isn`t supposed to.

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
marcelfloe: I can try, but i can't promise anything.

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
marcelfloe: I have no idea, what `var` is supposed to do and i don`t undestand, what you want to be changed here.

server/src/main/java/de/uftos/dto/solver/GradeProblemDto.java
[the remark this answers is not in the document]
danieldietzler: A record implicitly has a `id()` method. The interface should expect that method

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
marcelfloe: I guess that should work too.
I'll fix this.

server/src/main/java/de/uftos/dto/ucdl/ast/OperatorDto.java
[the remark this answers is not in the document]
danieldietzler: Oh cool!

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
marcelfloe: I'll think about it.

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
marcelfloe: Without cast it would be `addAll(resource.getStudents())` and `getStudents` will return a List.

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
marcelfloe: You mean the generics?
Yeah... They are fun...

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
marcelfloe: Thanks

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionTimefoldInstance.java
marcelfloe: I will fix that.

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/constraintinstances/ConstraintInstanceHardPenalize.java
marcelfloe: It looks sketchy, because it was...
Initially the Constraint-Provider (or EasyScoreCalculator) would fill the null-pointers of the lessons in, however they would check the constraints after each null-pointer got set and therefore those constraints would all throw null-pointer exceptions.
And it wasn't possible to check the relevant resources beforehand, as those relevant resources were unknown.
However by now the null values get pre-instantiated by the Repository instead of the Solver, so i should be able to simply remove these.

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/constraintinstances/ConstraintInstanceSoftReward.java
marcelfloe: I think i can actually collapse these down into one class and reduce code-duplication as we aren't using a ConstraintProvider anymore.

server/src/main/java/de/uftos/repositories/solver/timefold/domain/LessonTimefoldInstance.java
marcelfloe: Thats something only the timefold-devs know.

server/src/main/java/de/uftos/repositories/solver/timefold/domain/LessonTimefoldInstance.java
marcelfloe: I'll change it.

server/src/main/java/de/uftos/repositories/solver/timefold/domain/TimetableSolutionTimefoldInstance.java
marcelfloe: Will be added.

server/src/main/java/de/uftos/repositories/ucdl/parser/PredefinedConstraint.java
marcelfloe: I think an enum is easier to work with inside java as you don't have tons of "magic-Strings" everywhere.
I think a file would also work, but that would probably include a lot of implementation work.

server/src/main/java/de/uftos/dto/solver/GradeProblemDto.java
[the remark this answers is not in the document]
marcelfloe: I'll try to change that.

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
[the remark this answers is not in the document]
marcelfloe: Checkstyle asks for a default statement.

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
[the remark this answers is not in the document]
danieldietzler: Oh I see. That that's even a checkstyle is hilarious to me lmao. Although I guess it *is* a good indicator for methods that get too long *cough cough*

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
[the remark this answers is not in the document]
danieldietzler: Interesting, k

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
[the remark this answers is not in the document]
danieldietzler: So there could be a potential inconsistency against which you tried to harden the code, makes sense!

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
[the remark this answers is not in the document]
danieldietzler: Maybe you could update the comment to something like
> instantiates resources and their relations based on their ids
or something then :)

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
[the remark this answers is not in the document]
danieldietzler: The index part. `timefoldInstance.setIndex(lesson.index())`

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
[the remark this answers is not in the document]
danieldietzler: Oh, `ResourceTimefoldInstance` has a list of parameters as well? So we're talking about nested lists here?

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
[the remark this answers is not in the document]
danieldietzler: Ah got it. If they never have any side effects that's fine

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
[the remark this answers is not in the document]
danieldietzler: Yeaaaah... This is really confusing to me 😅

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
danieldietzler: I see. Then I may add a variable for keeping state or something. Idk.... just *anything* else to not have jump targets lol

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
danieldietzler: You can use `var` instead of explicitly defining a type. So for instance these two examples are exactly the same:
```java
var foo = "Hello World!"; // type String
String bar = "Hello World!"; // type String
```

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
danieldietzler: Oh sorry, I read it as `addAll(resource).getStudents()`

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/constraintinstances/ConstraintInstanceHardPenalize.java
danieldietzler: Nice!

server/src/main/java/de/uftos/repositories/solver/timefold/domain/LessonTimefoldInstance.java
danieldietzler: Then let's praise them 🙏🏼

server/src/main/java/de/uftos/repositories/ucdl/parser/PredefinedConstraint.java
danieldietzler: Ok, that works for me then :)

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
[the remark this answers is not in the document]
marcelfloe: Its getting much bigger refactoring

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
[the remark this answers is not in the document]
marcelfloe: All resources (including lessons), which got created only have their id and nothing else.
So all relations between the resources still need to be created.
Additionally there are further parameters for lessons (-> index) and timeslots (-> day, slot) which need to be set.
And here the index of the lessons is set based on the index of the dto.

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
marcelfloe: I guess thats doable.

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
marcelfloe: So do I understand it correctly, that you want me to replace `List<...>` with `var`?

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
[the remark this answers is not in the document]
danieldietzler: 👀

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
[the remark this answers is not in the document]
danieldietzler: Oh, it's like the actual lesson index everything uses. Not some specific solver index. Got it

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
danieldietzler: Yes!

marcelfloe: I honestly don't like that.
I prefer ugly but visible types instead of just writing `var` and leaving the rest to the compiler.

danieldietzler (approved):
Awesome!

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
```diff
@@ -1,25 +1,893 @@
...
+    for (ConstraintInstanceDto instance : timetable.instances()) {
+      constraintInstances.add(
+          ConstraintInstanceFactory.getConstraintInstance(instance, definitions, resources));
+    }
+    solution.getConstraintInstances().addAll(constraintInstances);
+
+    return solution;
+  }
+
+  TimetableSolutionDto getTimetableInstanceFromSolutionInstance(
```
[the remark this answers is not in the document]
marcelfloe: I thought i did add a "private" there...

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
```diff
@@ -1,25 +1,893 @@
...
+          )
+      );
+    }
+
+    return new TimetableSolutionDto(lessons);
+
+  }
+
+  @Override
+  public Future<TimetableSolutionDto> solve(TimetableProblemDto timetable)
```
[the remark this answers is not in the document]
marcelfloe: Should be doable.

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
```diff
@@ -1,25 +1,893 @@
...
+
+      LocalSearchPhaseConfig localSearch = new LocalSearchPhaseConfig();
+      localSearch.setAcceptorConfig(
+          new LocalSearchAcceptorConfig().withSimulatedAnnealingStartingTemperature(
+              "5hard/100soft"));
+
+
+      SolverConfig solverConfig = new SolverConfig()
+          .withTerminationConfig(new TerminationConfig()
+              .withMillisecondsSpentLimit(60000L))
```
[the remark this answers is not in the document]
marcelfloe: Yes there are (for example a feasable score), however it will require a lot of extra-effort to implement those.
Technically we need to scale everything based on the problem size...

server/src/main/java/de/uftos/repositories/solver/timefold/constraints/ConstraintDefinitionFactory.java
```diff
@@ -0,0 +1,1182 @@
...
+        Optional<Boolean> bool = function.apply(arguments);
+        if (bool.isPresent()) {
+          return bool.get();
+        }
+      }
+      return finalReturnValue.apply(arguments);
+    };
+  }
+
+  @SuppressWarnings("checkstyle:VariableDeclarationUsageDistance")
```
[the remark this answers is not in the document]
marcelfloe: Yes

PR #208, opened 2024-07-27
feat: ucdl consistency checks

danieldietzler:


server/src/main/java/de/uftos/services/UcdlEditorService.java
```diff
@@ -70,4 +74,169 @@ public Resource getUcdl() {
...
+  private boolean signatureChanged(ConstraintSignature signature,
+                                   ConstraintDefinitionDto definition) {
+    if (definition == null) {
+      return true;
+    }
+    if (!definition.name().equals(signature.getName())) {
+      throw new IllegalArgumentException();
+    }
+
+    Queue<ConstraintParameter> signatureParameters = new ArrayDeque<>(signature.getParameters());
```
danieldietzler, line 225: Using appropriate data structures, fancy!

danieldietzler (commented):
Server code looks good to me!

SuperMarcomen (approved):

PR #216, opened 2024-07-28
feat: beautify things

SuperMarcomen:


web/src/lib/elements/ui/dataTable/data-table.svelte
```diff
@@ -44,7 +44,7 @@
   export let additionalId: string = '';
   export let sortable = true;
   export let addButton = true;
-  export let pageSize = 15;
+  export let pageSize = 2;
```
danieldietzler, line 47: I don't think a default size of 2 is feasible

SuperMarcomen: You're right. I used that for testing but forgot removing it.

web/src/lib/elements/ui/dataTable/data-table.svelte
```diff
@@ -266,7 +268,7 @@
             >
               {#each row.cells as cell (cell.id)}
                 <Subscribe attrs={cell.attrs()} let:attrs>
-                  <Table.Cell {...attrs}>
+                  <Table.Cell class="mammt" {...attrs}>
```
danieldietzler, line 271: WTH is "mammt"?

danieldietzler (approved):

web/src/lib/elements/ui/dataTable/data-table.svelte
SuperMarcomen: I added a random class to see to which HTML element that cell would correspond and I forgot to remove it lol

PR #221, opened 2024-07-29
Feat/solver start

marcelfloe:


server/src/main/java/de/uftos/controller/TimetableController.java
```diff
@@ -42,7 +43,8 @@ public TimetableController(TimetableService timetablesService) {
    * @return the created timetable with the assigned ID.
    */
   @PostMapping()
-  public Timetable createTimetable(@RequestBody TimetableRequestDto timetable) {
+  public Timetable createTimetable(@RequestBody TimetableRequestDto timetable)
+      throws BadRequestException {
```
danieldietzler, line 47: The controller should not throw bad request exceptions.

server/src/main/java/de/uftos/services/TimetableService.java
```diff
@@ -60,9 +166,329 @@ public Timetable getById(String id) {
    * @throws ResponseStatusException is thrown if the ID defined in the timetable parameter is
    *                                 already present in the database.
    */
-  public Timetable create(TimetableRequestDto timetable) {
-    // solving magic here
-    return null;
+  public Timetable create(TimetableRequestDto timetable) throws BadRequestException {
```
danieldietzler, line 169: Again, services methods should only throw ResponseStatusExceptions, as they are being interpreted by spring with http Status codes

danieldietzler (changes requested):

jdiester (approved):
Looks good

danieldietzler (approved):

PR #254, opened 2024-08-05
feat: timetable service test

SuperMarcomen:


server/src/test/java/de/uftos/services/TimetableServiceTests.java
```diff
@@ -0,0 +1,232 @@
...
+
+  @Test
+  void nonExistingTimetableById() {
+    assertThrows(ResponseStatusException.class, () -> timetableService.getById("321"));
+  }
+
+  @Test
+  void createTimetable() {
+    TimetableRequestDto timetable = new TimetableRequestDto("2024");
+    assertDoesNotThrow(() -> timetableService.create(timetable));
```
danieldietzler, line 224: You should also expect that the repo method has been called

server/src/test/java/de/uftos/services/TimetableServiceTests.java
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  @Test
+  void timetableById() {
```
danieldietzler, line 228: `getTimetableById`

SuperMarcomen: done

server/src/test/java/de/uftos/services/TimetableServiceTests.java
```diff
@@ -0,0 +1,78 @@
...
+@MockitoSettings(strictness = Strictness.LENIENT)
+public class TimetableServiceTests {
+
+  @Mock
+  private TimetableRepository timetableRepository;
+  @Mock
+  private UcdlRepository ucdlRepository;
+
+  @Mock
+  private ServerRepository serverRepository;
```
jdiester, line 41: Are the server and ucdl repositories needed anywhere?

jdiester (approved):

server/src/test/java/de/uftos/services/TimetableServiceTests.java
SuperMarcomen: You are right, they are no longer used. I removed them

PR #265, opened 2024-08-06
Fixing problem with null pointers.

### LINKED ISSUE #251: Constraint definition with empty fields results in null-pointer-exception.
grafik
This constraint definition results in a null-pointer-exception and the backend doesn't return a ParsingResponse.

marcelfloe:
closes #251

### DIFF
--- a/server/src/main/java/de/uftos/repositories/ucdl/parser/UcdlParser.java
+++ b/server/src/main/java/de/uftos/repositories/ucdl/parser/UcdlParser.java
@@ -49,6 +49,7 @@ public class UcdlParser {
   }
 
   //parses one full definition
+  @SuppressWarnings("checkstyle:VariableDeclarationUsageDistance")
   private static ConstraintDefinitionDto parseConstraintDefinition(String name,
                                                                    JsonNode constraintDefinition)
       throws ParseException {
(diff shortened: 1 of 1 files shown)


server/src/main/java/de/uftos/repositories/ucdl/parser/UcdlParser.java
```diff
@@ -67,6 +68,14 @@ private static ConstraintDefinitionDto parseConstraintDefinition(String name,
 
     String description = constraintDefinition.get("description").textValue();
 
+    if (description == null) {
+      throw new ParseException("The description needs to contain a String!");
+    }
+
+    if (description.replaceAll("\\s", "").isEmpty()) {
+      throw new ParseException("The description requires a non-whitespace text!");
+    }
```
danieldietzler, line 77: Isn't there something like `isBlank`?

danieldietzler (approved):
Nice!

server/src/main/java/de/uftos/repositories/ucdl/parser/UcdlParser.java
danieldietzler: @marcelfloe ah yes https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/String.html#isBlank()

PR #270, opened 2024-08-07
Unit test for the server service

jdiester:


server/src/test/java/de/uftos/services/ServerServiceTests.java
```diff
@@ -0,0 +1,101 @@
...
+        () -> assertEquals(45, result.getTimeslotLength()),
+        () -> assertEquals(8, result.getTimeslotsAmount()),
+        () -> assertEquals("7:45", result.getStartTime()),
+        () -> assertEquals(0, result.getBreaks().length)
+    );
+  }
+
+  @Test
+  void setMetadata() {
+    TimetableMetadata newMetadata = new TimetableMetadata(90, 5, "8:00", new Break[] {});
```
marcelfloe, line 89: It should also be tested, whether changing the breaks works.

danieldietzler (approved):

PR #271, opened 2024-08-07
feat: fts for resources

danieldietzler:
# What this is doing
The core of this is relatively straightforward: 
We add a new column to resources we want to be able to search through. This column is of type `tsvector` and gets automatically generated based on all columns we want to be able to search across. Then, in the query we can have a `WHERE` statement that looks like this
```sql
WHERE <col_name> @@ websearch_to_tsquery('german', '<search string>')
```
For good measurement we slap a GIN index onto each column and we're off to the races.
Good resources to get started are https://supabase.com/blog/postgres-full-text-search-vs-the-rest followed by the official [docs](https://www.postgresql.org/docs/current/textsearch.html).

# Why is this so ugly then?
Unfortunately, Spring Data/Hibernate really blocks you from doing anything even remotely advanced. 
For starters, even getting the `tsvector` column type required some extra dependency. Then, non-default automatically generated columns aren't supported either. So, we can't just create a column and add `GENERATED ALWAYS AS to_tsvector(...) STORED`. We do however need the column to be in the java entity, as we'll see later. Thus, on startup (but *after* Spring Data initialized its stuff) a `schema.sql` file gets executed. Within that, we immediately delete the column that's not automatically generating values, and create a new one using raw SQL. (Side note: pg17 will allow to edit column types, so this part could become a bit less ugly)
Of course Spring Data/Hibernate doesn't support custom indices either (and they'd get deleted anyways), we also create those here. At this point, we have an automatically populating `tsvector` column for each resource table - cool.

Spring Data making stuff more complicated instead of easier seems to be a pattern here, so let's continue... Of course you can't just write raw SQL within a query builder (`Specification` in Spring-speak)~~, as that would be easy. We don't to that here~~. That's why I had to go the extra mile to create a custom function within the Spring ecosystem by using a custom "dialect". 
Now, we finally have everything together and we can write our `search` function in the `SpecificationBuilder` 🎉.

# TODO
- [x] Fix tests


server/src/main/java/de/uftos/services/RoomService.java
```diff
@@ -40,20 +40,16 @@ public RoomService(RoomRepository repository, ServerRepository serverRepository)
...
+   * @param search   the search filter.
+   * @param capacity the capacity filter.
+   * @param tags     the tags filter.
    * @return the page of entries fitting the parameters.
    */
-  public Page<Room> get(Pageable pageable, Optional<String> name,
-                        Optional<String> buildingName, Optional<Integer> capacity,
-                        Optional<String[]> tags) {
-    //currently no capacity filter
+  public Page<Room> get(Pageable pageable, Optional<String> search, Optional<Integer> capacity,
```
Lama64, line 49: Wouldn't it be better to remove the capacity parameter if it isn't used?
Why keep it, if you replace all the others?

server/src/main/java/de/uftos/utils/SpecificationBuilder.java
```diff
@@ -21,6 +21,7 @@ public class SpecificationBuilder<T> {
    * Constructs a new {@code SpecificationBuilder} with an empty specification.
    */
   public SpecificationBuilder() {
+    //specification = Specification.where(((root, query, cb) -> cb.disjunction()));
```
Lama64, line 24: Should be removed before merge, if it doesn't seem to work.

server/src/main/java/de/uftos/services/RoomService.java
danieldietzler: I don't think it makes sense to full text search for a number (open for other opinions though), however there may still be a scenario where you want to filter by capacity as well. In theory this could also become an additional field in the GUI, just having it available through the API certainly makes sense anyways though IMO.

server/src/main/resources/schema.sql
```diff
@@ -0,0 +1,41 @@
+ALTER TABLE teachers
```
marcelfloe, line 1: What is this file for?
When is the SQL used?

server/src/main/java/de/uftos/config/CustomPostgreSQLDialect.java
```diff
@@ -0,0 +1,20 @@
...
+    super();
+  }
+
+  @Override
+  public void initializeFunctionRegistry(FunctionContributions functionContributions) {
+    super.initializeFunctionRegistry(functionContributions);
+    var functionRegistry = functionContributions.getFunctionRegistry();
+    functionRegistry.registerPattern(
+        "tsvector_match",
+        "(?1 @@ ?2)"
```
marcelfloe, line 17: What does this line mean?

server/src/main/resources/schema.sql
danieldietzler: This is run right after the spring data initialization completes. What's this for is discussed in the description :)

marcelfloe:
Still broken:
grafik

server/src/main/java/de/uftos/config/CustomPostgreSQLDialect.java
danieldietzler: Again, as can be seen in the description, the actual search looks like `WHERE something @@ to_tsquery` and since you cannot write this in specifications ootb, I added a function that takes to arguments (1 and 2) and builds `?1 @@ ?2`

server/src/main/resources/schema.sql
marcelfloe: So it's run every time the initialization is completed?
This sounds a bit overkill...

danieldietzler: No, it's run every time something in the schema changes that spring considers worthwhile updating for. For prod this doesn't matter anyways, as it'll only run once on startup

server/src/main/java/de/uftos/config/CustomPostgreSQLDialect.java
```diff
@@ -0,0 +1,20 @@
+package de.uftos.config;
+
+import org.hibernate.boot.model.FunctionContributions;
+import org.hibernate.dialect.PostgreSQLDialect;
+
+public class CustomPostgreSQLDialect extends PostgreSQLDialect {
```
marcelfloe, line 6: JavaDoc missing.

marcelfloe (approved):

PR #279, opened 2024-08-09
Unit test for the tag service

jdiester:


server/src/test/java/de/uftos/services/TagServiceTests.java
```diff
@@ -0,0 +1,113 @@
...
+
+  @Mock
+  private ServerRepository serverRepository;
+
+  @InjectMocks
+  private TagService tagService;
+
+  @BeforeEach
+  void setUp() {
+    Tag tag = new Tag("", "Sehbehinderung");
```
danieldietzler, line 46: Minor nit: I personally prefer having tests (basically everything non user-facing) be completely in English.

danieldietzler (approved):

PR #264, opened 2024-08-06
Refactor datatable

Lama64:

PR #284, opened 2024-08-10
Fix: empty resources

Lama64:


danieldietzler (approved):
I think it makes sense if we provide proper error messages, since as of right now the user doesn't really now why exactly they got a 400 now

server/src/main/java/de/uftos/services/StudentService.java
```diff
@@ -67,10 +67,12 @@ public Student getById(String id) {
    *
    * @param student the student which is to be created.
    * @return the updated student which includes the ID that has been assigned.
-   * @throws ResponseStatusException is thrown if the ID defined in the student parameter
-   *                                 is already present in the database.
+   * @throws ResponseStatusException is thrown if the first or last name of the student is blank.
    */
   public Student create(StudentRequestDto student) {
+    if (student.firstName().isBlank() ||student.lastName().isBlank()) {
```
danieldietzler, line 73: Fix those formatting issues!

danieldietzler (changes requested):

danieldietzler (approved):

Lama64:
Wait for merge of #264

PR #313, opened 2024-08-15
Add hashCode

Lama64:


Lama64:
May not be possible to delete equals method.
Adding hashcodes may be necessary.

danieldietzler (approved):

Issue #238, opened 2024-08-05
Adding constraint instances adds those instances to all constraint definitions.


marcelfloe:
![grafik](https://github.com/user-attachments/assets/60ee11aa-d81d-4904-bdae-95c375d1e334)

marcelfloe:
This seems to be a problem only with the GUI. The backend seems to save the Instances correctly.

PR #338, opened 2024-08-20
Fixing dto creation for signature-instances.

marcelfloe:


marcelfloe:
closes #238

server/src/main/java/de/uftos/dto/responsedtos/ConstraintInstancesResponseDto.java
```diff
@@ -16,11 +16,13 @@
...
+ * @param parameters          the parameters of the constraint signature.
+ * @param totalElements       the total amount of instances of the signature.
  */
 public record ConstraintInstancesResponseDto(@NotNull List<SlimInstance> constraintInstances,
-                                             @NotNull
-                                             List<ConstraintArgumentDisplayName> displayNames,
-                                             @NotNull List<ConstraintParameter> parameters) {
+                                             @NotNull List<ConstraintArgumentDisplayName> displayNames,
+                                             @NotNull List<ConstraintParameter> parameters,
+                                             @NotNull long totalElements) {
```
danieldietzler, line 25: This should use `@IsPositiveOrZero` or something like this.

danieldietzler (approved):

danieldietzler (approved):

PR #339, opened 2024-08-20
Changed signature of POST of timetable endpoint.

marcelfloe:
Front-end changes required.


SuperMarcomen:
> Front-end changes required.

Done

danieldietzler (approved):

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
```diff
@@ -188,6 +189,13 @@ private TimetableSolutionTimefoldInstance getSolutionInstanceFromTimetableInstan
       constraintInstances.add(
           ConstraintInstanceFactory.getConstraintInstance(instance, definitions, resources));
     }
+    if (constraintInstances.isEmpty()) {
+      //solver won't work without constraint instances
```
danieldietzler, line 193: Should we even just throw an exception if having no constraint instances is nonsensical?

danieldietzler (approved):

server/src/main/java/de/uftos/repositories/solver/SolverRepositoryImpl.java
marcelfloe: I would not throw an error as it technically still is a valid input.
Also I like to use the solver without Constraints to fill the DB with lessen without needing to wait multiple minutes.

PR #343, opened 2024-08-21
Fix: removing multiple students / deleting multiple instances.

Lama64:


web/src/lib/elements/ui/dataTable/data-table-actions.svelte
```diff
@@ -7,7 +7,7 @@
   import { toast } from '$lib/utils/resources';
 
   export let id: string;
-  export let deleteEntry: (id: string, additionalId?: string) => Promise<void>;
+  export let deleteEntry: (id: string[], additionalId?: string) => Promise<void>;
```
danieldietzler, line 10: I would rename this, as it now deletes multiple entries.

web/src/lib/utils/resources.ts
```diff
@@ -253,28 +253,30 @@ export async function getStudentsFromGroup(
     throw error(400, { message: 'Invalid student group id' });
   }
   const studentGroup = await getStudentGroup(additionalId);
```
danieldietzler, line 255:
```suggestion
  const { students } = await getStudentGroup(additionalId);
```

web/src/lib/utils/resources.ts
```diff
@@ -253,28 +253,30 @@ export async function getStudentsFromGroup(
...
-  const dataItems: DataItem[] = studentGroup.students.map(
-    (student): DataItem => ({
-      id: student.id,
-      firstName: student.firstName,
-      lastName: student.lastName,
-      tags: student.tags.map(({ name }) => name),
-    }),
-  );
+  const dataItems: DataItem[] = studentGroup.students
```
danieldietzler, line 256: Use `(students || []).map((student) => )`

web/src/lib/utils/resources.ts
```diff
@@ -253,28 +253,30 @@ export async function getStudentsFromGroup(
...
+          firstName: student.firstName,
+          lastName: student.lastName,
+          tags: student.tags.map(({ name }) => name),
+        }),
+      )
+    : [];
   return {
     data: dataItems,
-    totalElements: Number(studentGroup.students.length),
+    totalElements: Number(studentGroup.students ? studentGroup.students.length : 0),
```
danieldietzler, line 268: Why does this need to be converted to a number? `studentGroup.students` should already be a number right? Also, you probably want `students?.length ?? 0`

web/src/lib/utils/resources.ts
```diff
@@ -30,6 +30,14 @@ export type DataItem = {
   [key: string]: string | string[] | number;
 };
 
+async function deleteMultiple(ids: string[], deleteEntry: (id: string) => Promise<void>) {
+  let promises: Promise<void>[] = [];
+  for (const id of ids) {
+    promises = [...promises, deleteEntry(id)];
+  }
+  return Promise.all(promises);
```
danieldietzler, line 38:
```suggestion
  return Promise.all(ids.map((id) => deleteEntry(id)));
```

danieldietzler (approved):

PR #347, opened 2024-08-21
API generation and missing javadoc

Lama64:


server/src/main/java/de/uftos/dto/responsedtos/StudentGroupResponseDto.java
```diff
@@ -5,22 +5,41 @@
...
+ * @param id       the id of the student group
+ * @param name     the name of the student group
+ * @param students the students in the student group
+ * @param grades   the grades of the student group
+ * @param tags     the tags of the student group
+ * @param lessons  the lessons of the student group
+ * @param subjects the subjects of the student group
+ */
+public record StudentGroupResponseDto(@NotNull String id,
+                                      @NotNull String name,
```
danieldietzler, line 24: Those should probably be `@NotEmpty`

danieldietzler (approved):

server/src/main/java/de/uftos/dto/responsedtos/StudentGroupResponseDto.java
Lama64: `NotNull` is used in every response dto.
If you change it here you could argue, that all others need to be changed as well.

Lama64: Nevermind im just blind

PR #349, opened 2024-08-21
feat: allow to force-update the ucdl file

### LINKED ISSUE #234: Allow for force-save of ucdl.yml
DSL-Editor doesn't allow to force-save new contents of the ucdl.yml-file yet.

### LINKED ISSUE #253: Upload in DSL view.
Uploading a .yml file in the DSL view results in the application trying to save the file immediately.
This requires an option to force-upload.
As an alternative it would also be possible to load the file into the view and save the contents from there.

### LINKED ISSUE #299: Confirm changes dialog even after save button was pressed.

danieldietzler:
Closes #234.
Closes #253.
Closes #299.

Also added an ugly workaround to use our custom modal instead of the browser alert when leaving the page with unsaved changes. If `beforeNavigate` would take an async function this could be nicely done with promises, but it doesn't so we need to get ugly :(

### DIFF
--- a/web/src/lib/components/editor.svelte
+++ b/web/src/lib/components/editor.svelte
@@ -1,23 +1,32 @@
 <script lang="ts">
-  import { beforeNavigate } from '$app/navigation';
+  import { beforeNavigate, goto } from '$app/navigation';
+  import * as AlertDialog from '$lib/elements/ui/alert-dialog';
   import Button from '$lib/elements/ui/button/button.svelte';
   import { getDefaultUcdlFile, getUcdlFile, setUcdlFile, validateUcdlFile } from '$lib/sdk/fetch-client';
+  import { toast } from '$lib/utils/resources';
   import { debounce } from 'lodash-es';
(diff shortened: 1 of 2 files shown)


web/src/lib/components/editor.svelte
```diff
@@ -132,3 +181,44 @@
...
+  <AlertDialog.Content>
+    <AlertDialog.Header>
+      <AlertDialog.Title>Datei konnte nicht gespeichert werden</AlertDialog.Title>
+      <AlertDialog.Description>"{saveAlertMessage}" Trotzdem speichern?</AlertDialog.Description>
+    </AlertDialog.Header>
+    <AlertDialog.Footer>
+      <AlertDialog.Cancel class="shadow-custom">Nein</AlertDialog.Cancel>
+      <AlertDialog.Action
+        on:click={async () => {
+          saveAlertOpen = false;
```
Lama64, line 196: This is not needed, the alert closes automatically.
```suggestion
```

Lama64 (approved):

PR #352, opened 2024-08-22
fix: navbar settings button

SuperMarcomen:


web/src/lib/elements/ui/navbar/navbar.svelte
```diff
@@ -1,22 +1,80 @@
...
+        </Button>
+      </DropdownMenu.Trigger>
+      <DropdownMenu.Content class="w-56">
+        <DropdownMenu.Group>
+          {#each settings as setting}
+            <DropdownMenu.Item
+              class="text-lg"
+              on:click={() => {
+                window.location.href = setting.url;
+              }}>{setting.label}</DropdownMenu.Item
```
danieldietzler, line 63: `on:click={() => (window.location.href = setting.url)}`

web/src/lib/elements/ui/navbar/navbar.svelte
```diff
@@ -1,22 +1,80 @@
...
+            >
+          {/each}
+          <DropdownMenu.Item
+            class="text-lg"
+            on:click={async () => {
+              const name = new Date().getFullYear() + ':' + Date.now();
```
danieldietzler, line 69:
```suggestion
              const name = `${new Date().getFullYear()}:${Date.now()}`;
```

danieldietzler (approved):

web/src/lib/elements/ui/navbar/navbar.svelte
SuperMarcomen: done

PR #357, opened 2024-08-22
fix: delete used entities

SuperMarcomen:
It is now possible to delete an entity even if it is referenced somewhere else


server/src/main/java/de/uftos/repositories/database/StudentGroupRepository.java
```diff
@@ -12,4 +16,10 @@
 public interface StudentGroupRepository
     extends ListPagingAndSortingRepository<StudentGroup, String>,
     ListCrudRepository<StudentGroup, String>, JpaSpecificationExecutor<StudentGroup> {
+
+  List<StudentGroup> findByStudents(Student student);
+
+  List<StudentGroup> findByTags(Tag tag);
+
+  List<StudentGroup> findBySubjects(Subject subject);
 }
```
marcelfloe, line 25:
```suggestion
}

```

marcelfloe (approved):

PR #358, opened 2024-08-23
Solver tests

marcelfloe:


jdiester (approved):
Didn't understand everything, but everything I did understand looks good

PR #363, opened 2024-08-23
chore: persist timeslots when updating timetable metadata

### LINKED ISSUE #295: Timeslots can't be used in constraint instances.

### LINKED ISSUE #359: .init function of UftosApplication does not save the created timeslots in the timelots relation

danieldietzler:
Closes #295
Closes #359


server/src/main/java/de/uftos/services/ServerService.java
```diff
@@ -77,6 +93,42 @@ public void setTimetableMetadata(TimetableMetadata timetableMetadata) {
...
+          (slot) -> slot.getSlot() < currentTimeslots.size() / Weekday.values().length);
+
+      this.timeslotRepository.saveAll(timeslots);
+      return;
+    }
+
+    List<Timetable> timetables = this.timeslotRepository.findAll().stream()
+        .flatMap(timeslot -> timeslot.getLessons().stream().map(Lesson::getTimetable)).distinct()
+        .toList();
+    this.timeslotRepository.deleteAll();
```
marcelfloe, line 127: Deleting all timeslots will result in all constraint instances being lost as the ids of the new timeslots are not the same as the ids of the old timeslots.
Only removed timeslots (and their instances) should be deleted.

danieldietzler: Hm true that sounds correct

danieldietzler: Should be fixed.

server/src/main/java/de/uftos/services/ServerService.java
```diff
@@ -77,6 +93,43 @@ public void setTimetableMetadata(TimetableMetadata timetableMetadata) {
...
+
+    List<Timetable> timetables = currentSlots.stream()
+        .flatMap(timeslot -> timeslot.getLessons().stream().map(Lesson::getTimetable)).distinct()
+        .toList();
+    this.timeslotRepository.deleteAll(currentSlots);
+    this.timetableRepository.deleteAll(timetables);
+
+    // TODO delete instances that used timeslots that don't exist anymore
+
+    this.timeslotRepository.saveAll(timeslots);
```
marcelfloe, line 133: What is this line for?
New timeslots are already saved in line 119.
Saving here would create new timeslots for already exisiting timeslots as the created timeslots don`t have ids.
And as there are no fields of the timeslots which require updating I think this line should simply be removed.

danieldietzler: Oh yeah missed that, good catch!

marcelfloe:
closes #295 and #359

SuperMarcomen (approved):
LGTM

PR #377, opened 2024-08-26
Fix: delete multiple entities

SuperMarcomen:


server/src/main/java/de/uftos/services/StudentGroupService.java
```diff
@@ -145,11 +145,10 @@ public StudentGroupResponseDto update(String id, StudentGroupRequestDto groupReq
 
     List<Grade> grades =
         this.gradeRepository.findAllById(groupRequest.gradeIds());
-    List<Grade> allGrades = this.gradeRepository.findAll();
-    for (Grade grade : allGrades) {
-      grade.getStudentGroups().remove(group);
-    }
     for (Grade grade : grades) {
+      if (grade.getStudentGroups().contains(group)) {
```
Lama64, line 149: What about the grades that are no longer present in the `groupRequest`, but where previously referencing the group. These now reference a grade that they shouldn't.

You could just get the old group saved in the db before replacing it and remove the group from all grades in  the difference of the new and old groups grades.

Why use an if with a continue and not just `!(grade.getStudentGroups().contains(group))` as the condition and the add and save in the body?

marcelfloe (commented):
This seems to be missing rooms, grades, teachers and tags.

server/src/main/java/de/uftos/services/StudentGroupService.java
```diff
@@ -145,11 +145,10 @@ public StudentGroupResponseDto update(String id, StudentGroupRequestDto groupReq
 
     List<Grade> grades =
         this.gradeRepository.findAllById(groupRequest.gradeIds());
-    List<Grade> allGrades = this.gradeRepository.findAll();
-    for (Grade grade : allGrades) {
-      grade.getStudentGroups().remove(group);
-    }
```
marcelfloe, line 151: Why is this loop thrown out?

server/src/main/java/de/uftos/services/StudentService.java
```diff
@@ -114,17 +101,7 @@ public Student update(String id, StudentRequestDto studentRequest) {
     }
     Student student = studentRequest.map();
     student.setId(id);
-    this.repository.save(student);
-
-    List<StudentGroup> studentGroups =
```
marcelfloe, line 119: I guess this is thrown out as the relation between student groups and students is handled at the student group endpoint.

server/src/test/java/de/uftos/services/SubjectServiceTests.java
```diff
@@ -143,4 +143,6 @@ void deleteExistentSubject() {
   void deleteNonExistentSubject() {
     assertThrows(ResponseStatusException.class, () -> subjectService.delete("nonExistentId"));
   }
+
+
```
marcelfloe, line 147: These new lines should be removed.

marcelfloe (commented):
Java-Part looks mostly good so far, only minor problems.
But I'm unable to verify anything inside the web-folder (+ uftos-openapi-specs.json).

server/src/main/java/de/uftos/services/StudentGroupService.java
SuperMarcomen: You are right, I forgot to do that.

I personally prefer early returns, but in this case in doesn't make that big of a difference

server/src/main/java/de/uftos/services/StudentGroupService.java
SuperMarcomen: Because I replaced it with a more efficient version down below (I forgot a delete statement though. I will fix it)

server/src/main/java/de/uftos/services/StudentService.java
SuperMarcomen: yes

marcelfloe:
Deleting grades removes them from student groups and leaves those student groups invalid.

server/src/main/java/de/uftos/services/GradeService.java
```diff
@@ -139,9 +144,15 @@ public GradeResponseDto update(String id, GradeRequestDto gradeRequest) {
    * @throws ResponseStatusException is thrown if no grade exists with the given ID.
    */
   public void delete(String id) {
-    var grade = this.repository.findById(id);
+    Optional<Grade> grade = this.repository.findById(id);
     if (grade.isEmpty()) {
-      throw new ResponseStatusException(HttpStatus.BAD_REQUEST);
+      throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
+          "Could not found a grade with this id");
```
marcelfloe, line 150:
```suggestion
          "Could not find a grade with this id!");
```

marcelfloe (commented):
Changes look good.

server/src/main/java/de/uftos/services/RoomService.java
```diff
@@ -107,7 +108,8 @@ public Room create(RoomRequestDto room) {
    * @throws ResponseStatusException is thrown if the name, building name are blank or the capacity is 0.
    */
   public Room update(String id, RoomRequestDto roomRequest) {
-    if (roomRequest.name().isBlank() || roomRequest.buildingName().isBlank() || roomRequest.capacity() == 0) {
+    if (roomRequest.name().isBlank() || roomRequest.buildingName().isBlank() ||
```
Lama64, line 111: Checkstyle

Lama64 (approved):

PR #391, opened 2024-08-27
feat: make timetable work

danieldietzler:


marcelfloe:
We also need timetables for individual students as those are part of the mandatory criteria.

marcelfloe:
grafik
It seems like the timetable display does not display multiple days.
Also the resource collection drop down quite liturally drops down on the page once a resource is selected.
Trying to change the resource after you once selected a resource of the same type results in an error page:
grafik
To successfully select another resource you first have to change the resource type (e.g. teacher -> room) and than change the resource type back to be able to have a new "first" selection of that resource type.

marcelfloe:
Also the disply for student groups is currently broken, but this is supposed to be due to an issue which is supposed to be fixed in #381

marcelfloe:
Student group timetable is no longer more broken than the others.
It seems like the timetable actually displays the lessons across multiple day:
grafik
(image is zoomed out to fit the entire timetable into the image)
But the white backdrop doesn't adjust for the correct size.

Also the timetables for multiple resources can't be shown and throw a BadRequestException instead:
grafik
(This could also be a problem with invalid timetable data: e.g. collisions betwenn multiple lessons during the same timeslot.
But I can't guarantee this.
 Also as the admin isn't technically required to choose anti-collision constraints for the resources the timetable display should probably be able to somehow display such collisions.
But everything below the last image is pure speculation on my part)

marcelfloe:
Also: The times on the left of the timetable don't adjust acording to the settings.

web/src/routes/timetable/[type]/[[id]]/+page.svelte
```diff
@@ -0,0 +1,157 @@
...
+
+  export let data;
+  let items: ComboBoxItem[] = [];
+  let selectedId: string = '';
+
+  $: id = data.id;
+  $: selectedId = items.map(({ value }) => value).includes(selectedId) ? selectedId : items.at(0)?.value || '';
+
+  const onTimetableSearch = (value: string) => {
+    items = items.filter((item) => item.label.toLowerCase().includes(value.toLowerCase()));
```
Comment (Lama64) [inline web/src/routes/timetable/[type]/[[id]]/+page.svelte:35]: Wouldn't this result in the search only working correctly one time and removing the search parameter wouldn't display all items again, as `items` gets overwritten and thus loosing the information about all items that were removed.

Comment (danieldietzler) [reply web/src/routes/timetable/[type]/[[id]]/+page.svelte:35]: Oh I think you are correct. This does sound weird

Comment (danieldietzler) [reply web/src/routes/timetable/[type]/[[id]]/+page.svelte]: Fixed, thanks!

marcelfloe:
grafik
This is the result in the DB after trying to delete a student group.
The student group only lost its relation to grades, but didn't get deleted.

Lama64 (approved):

PR #416, opened 2024-08-31
feat: error messages timetable generation

SuperMarcomen:
…metable whit missing entities

### DIFF
--- a/server/src/main/java/de/uftos/services/TimetableService.java
+++ b/server/src/main/java/de/uftos/services/TimetableService.java
@@ -197,9 +197,8 @@ public class TimetableService {
     timetableRepository.save(timetableEntity);
     try {
       TimetableProblemDto problemInstance = getProblemInstance(timetableEntity);
-      solverRepository.solve(problemInstance, solverFinishedEvent).get();
-    } catch (InterruptedException | ExecutionException | BadRequestException
-             | ResponseStatusException e) {
+      solverRepository.solve(problemInstance, solverFinishedEvent);
+    } catch (BadRequestException | ResponseStatusException e) {
       return new SuccessResponse(false, e.getMessage());
     }
     return new SuccessResponse(true, "Solver erfolgreich gestartet!");
--- a/server/src/main/java/de/uftos/services/TimetableService.java
+++ b/server/src/main/java/de/uftos/services/TimetableService.java
@@ -288,6 +288,33 @@ public class TimetableService {
       throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
           "Es muss mindestens eine Schülergruppe vorhanden sein, um einen Stundenplan zu erstellen");
     }
+
+    List<Teacher> teachers = teacherRepository.findAll();
+    if (teachers.isEmpty()) {
+      throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
+          "Es muss mindestens ein Lehrer vorhanden sein, um einen Stundenplan zu erstellen");
+    }
+
+    List<Room> rooms = roomRepository.findAll();
+    if (rooms.isEmpty()) {
+      throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
+          "Es muss mindestens ein Raum vorhanden sein, um einen Stundenplan zu erstellen");
+    }
+
+    List<Timeslot> timeslots = timeslotRepository.findAll();
+    if (timeslots.isEmpty()) {
+      throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
+          "Es muss mindestens ein Timeslot vorhanden sein, um einen Fahrplan zu erstellen");
+    }
+
+    List<Server> serverData = serverRepository.findAll();
+    // This shouldn't be possible since it should always contain at least one element,
+    // but you never know...
+    if (serverData.isEmpty()) {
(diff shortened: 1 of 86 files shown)


marcelfloe (commented):
Edit: this comment was nonsense...

server/src/main/java/de/uftos/services/TimetableService.java
```diff
@@ -276,28 +278,74 @@ private List<GradeProblemDto> getGrades() {
   private List<LessonProblemDto> getLessons(Timetable timetable) {
     //initializing new Lessons
     List<Curriculum> curriculums = curriculumRepository.findAll();
-    for (StudentGroup studentGroup : studentGroupRepository.findAll()) {
+    if (curriculums.isEmpty()) {
```
marcelfloe, line 281: I don't think the checks for curricula/lesson-counts/student-groups are necessary.

marcelfloe (commented):
I belief, all error messages (except for the server-message and the once I suggested to remove) should already work (when creating a timetable with a valid curriculum and a student group; without the solver will simply return an empty timetable) as those should already be catched by the solver and forwarded using BadRequestExceptions

Edit: Nvm...
Those are IllegalArgumentExceptions.
But if we catch those as well, then the extra exceptions should be redundant

server/src/main/java/de/uftos/services/TimetableService.java
```diff
@@ -276,28 +278,74 @@ private List<GradeProblemDto> getGrades() {
...
+    }
+    for (StudentGroup studentGroup : studentGroups) {
       for (Curriculum curriculum : curriculums) {
         if (curriculum.getGrade() != studentGroup.getGrades().getFirst()) {
           continue;
         }
-        for (LessonsCount lessonsCount : curriculum.getLessonsCounts()) {
+
+        List<LessonsCount> lessonsCounts = curriculum.getLessonsCounts();
+        // TODO is this needed?
```
marcelfloe, line 298:
```suggestion
```

server/src/main/java/de/uftos/services/TimetableService.java
```diff
@@ -276,28 +278,74 @@ private List<GradeProblemDto> getGrades() {
...
+        if (lessonsCounts.isEmpty()) {
+          throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
+              "There must be at least one lesson set in a curriculum to create a timetable");
```
marcelfloe, line 301: This should probably say something like "There must be at least on existing subject!"
Although this will not prevent, that all lesson counts are 0

marcelfloe (changes requested):
The error-messages should be in german

server/src/main/java/de/uftos/services/TimetableService.java
```diff
@@ -360,7 +407,12 @@ private List<StudentGroupProblemDto> getStudentGroups(Timetable timetable) {
 
   private List<StudentProblemDto> getStudents() {
     List<StudentProblemDto> students = new ArrayList<>();
-    for (Student student : studentRepository.findAll()) {
+    List<Student> studentList = studentRepository.findAll();
+    if (studentList.isEmpty()) {
+      throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
+          "There must be at least one student to create a timetable");
```
marcelfloe, line 413: Translation required.
Also I'm not so sure about this one...
The timetable has no problems with empty student groups.
I think this one should actually be removed.

marcelfloe (changes requested):

server/src/main/java/de/uftos/services/TimetableService.java
SuperMarcomen: Removed

server/src/main/java/de/uftos/services/TimetableService.java
```diff
@@ -376,7 +428,12 @@ private List<StudentProblemDto> getStudents() {
 
   private List<SubjectProblemDto> getSubjects(Timetable timetable) {
     List<SubjectProblemDto> subjects = new ArrayList<>();
-    for (Subject subject : subjectRepository.findAll()) {
+    List<Subject> subjectList = subjectRepository.findAll();
+    if (subjectList.isEmpty()) {
+      throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
+          "There must be at least one subject to create a timetable");
```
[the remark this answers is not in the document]
SuperMarcomen: I translated it

server/src/main/java/de/uftos/services/TimetableService.java
```diff
@@ -376,7 +423,12 @@ private List<StudentProblemDto> getStudents() {
...
```
marcelfloe, line 429: Still requires translation.

marcelfloe (changes requested):

server/src/main/java/de/uftos/services/TimetableService.java
marcelfloe:
```suggestion
          "Es muss mindestens ein Fach vorhanden sein, um einen Stundenplan zu erstellen");
```

marcelfloe:
Some checks seem like they do not work, as starting the solver only with a grade (and therefore also a curriculum), one student group and one subject results in an IllegalArgumentException.

marcelfloe (approved):

PR #417, opened 2024-08-31
feat: grade delete error toast

SuperMarcomen:


server/src/main/java/de/uftos/services/GradeService.java
```diff
@@ -157,25 +158,25 @@ public GradeResponseDto update(String id, GradeRequestDto gradeRequest) {
    * @param ids the IDs of the grades which are to be deleted.
    * @throws ResponseStatusException is thrown if no grade exists with the given ID.
    */
-  public void deleteGrades(String[] ids) {
+  public SuccessResponse deleteGrades(String[] ids) {
     List<String> gradesIds = Arrays.asList(ids);
     List<Grade> grades = this.repository.findAllById(gradesIds);
     if (grades.size() != gradesIds.size()) {
       throw new ResponseStatusException(HttpStatus.BAD_REQUEST,
```
marcelfloe, line 165: This should be replaced by a SuccessResponse.

marcelfloe (changes requested):

server/src/main/java/de/uftos/services/GradeService.java
SuperMarcomen: I cannot do it, otherwhise a test will fail, since it excepts an exception to be thrown. And I'm tired of modifying tests.

marcelfloe (approved):

PR #419, opened 2024-08-31
Feat/expanded timetable service tests

marcelfloe:


marcelfloe:
Some checks seem like they do not work, as starting the solver only with a grade (and therefore also a curriculum), one student group and one subject results in an IllegalArgumentException.

jdiester (commented):
There are some checkstyle issues

jdiester (approved):
Looks good!

